PR #558, opened 2026-07-23
Narrow RH5109/RH7101 code fix formatting scope to the diagnostic

thoenissen:
## Summary

Narrow the RH5109 and RH7101 code fixes so each only edits the syntax it actually changes, instead of reformatting the whole owning member or type.

## Why

Both fixes delegated an oversized scope to the formatter: RH5109 rebuilt the parameter list one-per-line and then ran the pipeline over the entire owning member, and RH7101 delegated the whole containing type to the full pipeline. The resulting fix diffs contained unrelated whole-member/whole-type reformatting and exposed the fixes to body/type-scope formatter defects.

## Linked issues

Closes #456

## Review notes

- RH5109 now returns the manually rebuilt, aligned parameter-list text directly. The formatter call was redundant for the guarded scope (the registration already rejects parameter lists carrying comments or directives), and formatting an isolated parameter list mis-aligned its continuation lines.
- RH7101 now runs only the field-split structural transform over the owning type instead of the full pipeline. The 1→N split still forces a type-level node, but nothing else in the type is touched.
- `FieldDeclarationSplitTransform` now derives member indentation from the leading whitespace when the field is the first member (whose leading trivia has no end-of-line — the newline sits on the opening brace), so the split indents generated siblings correctly on its own. Full-pipeline output is unchanged, since the later indentation phase already produced this result.

## Follow-up work

None. The underlying formatter defects the oversized scopes exposed are tracked separately; this change stops the fixes from reaching them.


Reihitsu.Analyzer.CodeFixes/Rules/Organization/RH7101DoNotCombineFieldsCodeFixProvider.cs
```diff
@@ -37,9 +38,23 @@ private static async Task<Document> ApplyCodeFixAsync(Document document, FieldDe
...
+        {
+            return document;
+        }
+
+        // The 1→N replacement the split needs forces a type-level scope, but running the full formatting pipeline
+        // over the type reformats members unrelated to the diagnostic. Reuse the formatter's field-split structural
+        // transform on its own instead: it splits the combined fields and preserves the comments attached to
+        // declarators and their separators, without touching the rest of the type.
+        var context = new FormattingContext(ReihitsuFormatterHelpers.DetectEndOfLine(root));
+        var splitTypeDeclaration = new FieldDeclarationSplitTransform(context, cancellationToken).Visit(typeDeclaration);
```
thoenissen, line 53: The split transform refuses directive-bearing fields (`CarriesDirective`, `FieldDeclarationSplitTransform.cs:200`), but `RegisterCodeFixesAsync` does not skip them, so this split-only path is a guaranteed no-op on a combined field carrying a preprocessor directive: it leaves the field intact, the document text is unchanged, and the RH7101 diagnostic survives after the fix is applied. Guard the registration with the same directive check so the fix is not offered when it cannot converge.

```csharp
internal class C
{
#if DEBUG
    private int a, b;   // RH7101 offered here; applying the fix changes nothing
#endif
}
```

---

thoenissen (commented):
Scope narrowing is correct: RH5109 returns the rebuilt parameter-list text directly and RH7101 runs only the field-split transform over the type; both new tests confirm surrounding members are untouched, and the `GetMemberIndentationTrivia` fallback correctly indents first-member siblings in the split-only path. Issue #456 requirements are covered.

One finding is inline (RH7101 fix is a no-op on directive-bearing combined fields).

---

Reihitsu.Analyzer.CodeFixes/Rules/Organization/RH7101DoNotCombineFieldsCodeFixProvider.cs
thoenissen: Addressed: `RegisterCodeFixesAsync` now skips combined fields that carry a preprocessor directive, reusing the split transform's own `CarriesDirective` check (exposed as `internal`), so the fix is no longer offered where it cannot converge. Added a regression test asserting no action is registered for a directive-bearing combined field (111477a).

---

thoenissen: Verified fixed: `RegisterCodeFixesAsync` now skips fields where `CarriesDirective` is true, mirroring the split transform's guard, so the fix is no longer offered on a directive-bearing combined field. The regression test defines `DEBUG` so the field is active and asserts no action is registered. Resolving.

---

PR #562, opened 2026-07-24
Fix RH2004 code fix inserting a conflicting modifier on partial types

thoenissen:
## Summary

Fix the `RH2004` code fix so it no longer inserts a conflicting accessibility modifier on partial types (CS0262). For partial type declarations the fix now reads the accessibility from the merged declared symbol instead of picking it syntactically, mirroring the sibling `RH7104` fix.

## Why

`RH2004AccessModifierMustBeDeclaredCodeFixProvider` chose the modifier syntactically — `internal` at namespace level, `private` when nested. For a routine split-class shape such as `partial class C` whose other part declares `public partial class C`, it inserted `internal` next to the `public` part, so the fixed code no longer compiled (CS0262).

## Linked issues

Closes #458

## Review notes

- For a partial type the fix resolves accessibility from `GetDeclaredSymbol(...).DeclaredAccessibility`, which merges all partial parts; non-partial declarations keep the existing syntactic default, so behavior for them is unchanged.
- A new `DeclarationModifierUtilities.AddAccessibilityModifiers(..., IReadOnlyList<SyntaxKind>)` overload supports compound accessibilities (`protected internal`, `private protected`). The existing single-keyword overloads now delegate to it, so their behavior is preserved (covered by the existing Core tests).
- Regression tests reproduce the CS0262 case (two-part partial adopting `public`) and the compound case (nested partial adopting `protected internal`), plus guards that single-part partials keep the `internal`/`private` defaults.

## Follow-up work

None. The `RH2004`/`RH7104` overlap and the `file`-local-types concern noted in the issue are tracked separately.


Reihitsu.Analyzer.CodeFixes/Rules/Design/RH2004AccessModifierMustBeDeclaredCodeFixProvider.cs
```diff
@@ -39,14 +40,60 @@ private static async Task<Document> ApplyCodeFixAsync(Document document, MemberD
...
-        return document.WithSyntaxRoot(updatedRoot);
+        return DeclarationModifierUtilities.AddAccessibilityModifier(memberDeclaration, defaultModifier);
+    }
+
+    /// <summary>
+    /// Maps a declared accessibility to the modifier keywords that express it
+    /// </summary>
+    /// <param name="accessibility">Declared accessibility</param>
+    /// <returns>The modifier keywords in declaration order</returns>
+    private static IReadOnlyList<SyntaxKind> GetAccessibilityModifiers(Accessibility accessibility)
```
thoenissen: `GetAccessibilityModifiers` duplicates the `Accessibility` → modifier-keyword mapping already implemented in `RH7104PartialElementsMustDeclareAccessModifierCodeFixProvider.GetAccessibilityTokens` (identical `Public`/`Internal`/`Protected`/`Private`/`ProtectedOrInternal`/`ProtectedAndInternal` cases, including the `private protected` and `protected internal` keyword orders). Extract the mapping into `Reihitsu.Core` (for example `DeclarationModifierUtilities`) and have both code fixes consume it, so the accessibility-keyword policy lives in one place.

---

thoenissen (commented):
Reviewed the RH2004 partial-type accessibility fix. The core change is correct: reading `DeclaredAccessibility` from the merged partial symbol resolves the CS0262 case, the trivia handling in the shared `AddAccessibilityModifiers` overload is preserved, and single-part partials still fall back to the `internal`/`private` defaults. One finding (medium, duplication) is posted inline — the `Accessibility` → keyword mapping should be consolidated with RH7104's copy into `Reihitsu.Core`.

---

Reihitsu.Analyzer.CodeFixes/Rules/Design/RH2004AccessModifierMustBeDeclaredCodeFixProvider.cs
thoenissen: Addressed in 1cb4dca. Extracted the `Accessibility` → keyword mapping into `DeclarationModifierUtilities.GetAccessibilityModifierKinds` (single canonical order for `protected internal` / `private protected`); both `RH2004` and `RH7104.GetAccessibilityTokens` now consume it. `NotApplicable` returns an empty list, so each caller keeps its own context-aware fallback — `RH2004` falls through to its nesting-aware syntactic default and `RH7104` keeps nested-`private` / namespace-`internal`. Added focused Core tests for the mapping.

---

thoenissen: Confirmed fixed in 1cb4dca. The `Accessibility` → keyword mapping now lives once in `DeclarationModifierUtilities.GetAccessibilityModifierKinds` (canonical `protected internal` / `private protected` order), and both `RH2004` and `RH7104.GetAccessibilityTokens` consume it, each keeping its own context-aware fallback for the empty `NotApplicable` result. Duplication is gone; resolving.

---

PR #563, opened 2026-07-25
Guard the RH7309 region swap against relocating preprocessor directives

thoenissen:
## Summary

`RH7309RegionsShouldFollowCategoryOrderCodeFixProvider` no longer offers a fix when exchanging the region blocks would change what the surrounding preprocessor directives mean. `CanReorderSafely` inspects every swapped region block and every stretch of text between two blocks, and refuses on either of two conditions, backed by new predicates in `Reihitsu.Core`:

- `SyntaxTriviaUtilities.ContainsUnbalancedConditionalDirectives` — a conditional directive (`#if`/`#elif`/`#else`/`#endif`) whose partner lies outside the segment.
- `SyntaxTriviaUtilities.ContainsPositionSensitiveDirectives` — any directive that is neither a region nor a conditional directive, covering `#pragma warning`, `#nullable` and `#line`.

## Why

The reorder exchanged whole `#region … #endregion` text blocks with no directive guard, which broke two distinct ways on source that compiles cleanly today:

- An `#if … #endif` wrapping one of the two regions but not the other: the swap moves the wrapped members out of the conditional section and the unwrapped members into it.
- A `#pragma warning disable` (or `#nullable`, `#line`) inside a region or in the gap between regions: these apply from where they sit onwards, so relocating the surrounding block silently changes their coverage. A moved `#pragma warning disable CS0169` stops suppressing the warning it was added for — a build break under `TreatWarningsAsErrors`.

Worth noting: the repro in #461 (an `#if` opening inside one region and closing inside another) is not valid C#. Roslyn refuses an `#endregion` that closes a region opened outside the enclosing `#if`, so that source already carries `CS1038`/`CS1027` and parses the directives as `BadDirectiveTrivia`, leaving `GetTopLevelRegions` with nothing to reorder. It is covered regardless — fewer than two resolvable regions also refuses, so that input no longer gets a no-op code action.

## Linked issues

Closes #461

## Review notes

- Directives that keep their meaning still get the fix, each with a test: a conditional block wholly inside one region, wholly between two regions, or wrapping *all* reordered regions, and any directive sitting ahead of every reordered region.
- `ContainsPositionSensitiveDirectives` refuses on *any* unrecognised directive kind rather than matching an enumerated list, so it fails closed for kinds it does not know about, including malformed ones. `#region`/`#endregion` are excluded deliberately: they carry no compilation meaning, and a region reorder moves them on purpose.
- Both Core predicates fail closed on a `null` root — an uninspectable span refuses the rewrite instead of permitting it.
- The guard sits in `RegisterCodeFixesAsync`, so no action is offered rather than a no-op action; `ReorderRegionsAsync` repeats the check, covering the `BatchFixer` fix-all path. The per-type decision is memoized so the region walk runs once per type, not once per diagnostic.
- `documentation/rules/RH7309.md` documents both withheld cases and the shapes that still get the fix.

## Follow-up work

- A reorder guard that needs `#region`/`#endregion` treated as unsafe — the way `OrderingDeclarationUtilities.MoveRangeContainsDirectives` treats every directive — should get its own predicate (for example `ContainsAnyDirective`) rather than a flag on `ContainsPositionSensitiveDirectives`, which answers a different question.

### DIFF
--- a/Reihitsu.Analyzer.CodeFixes/Rules/Organization/RH7309RegionsShouldFollowCategoryOrderCodeFixProvider.cs
+++ b/Reihitsu.Analyzer.CodeFixes/Rules/Organization/RH7309RegionsShouldFollowCategoryOrderCodeFixProvider.cs
@@ -87,11 +87,12 @@ public class RH7309RegionsShouldFollowCategoryOrderCodeFixProvider : CodeFixProv
     }
 
     /// <summary>
-    /// Determines whether the top-level regions can be exchanged without breaking conditional compilation.
-    /// The fix swaps whole region texts, so a conditional directive that is not paired inside the block it
-    /// belongs to — an <c>#if</c> that wraps one region but not the other, or one that opens inside a region
-    /// and closes outside it — would be separated from its partner. Depending on the shape that either moves
-    /// members into or out of a conditional section or leaves the document with unmatched directives
+    /// Determines whether the top-level regions can be exchanged without changing what the preprocessor
+    /// directives around them mean. The fix swaps whole region texts, which is unsafe in two ways: a
+    /// conditional directive that is not paired inside the block it belongs to gets separated from its partner,
+    /// and a position sensitive directive such as <c>#pragma warning</c>, <c>#nullable</c> or <c>#line</c> ends
+    /// up covering different code than before. Both are checked for every swapped block and for the text
+    /// between two blocks
     /// </summary>
     /// <param name="typeDeclaration">Type declaration whose regions should be reordered</param>
     /// <param name="regions">Top-level region pairs of the type</param>
@@ -112,12 +113,12 @@ public class RH7309RegionsShouldFollowCategoryOrderCodeFixProvider : CodeFixProv
 
             if (previousBlockEnd >= 0
                 && previousBlockEnd < span.Start
-                && SyntaxTriviaUtilities.ContainsUnbalancedConditionalDirectives(typeDeclaration, TextSpan.FromBounds(previousBlockEnd, span.Start)))
+                && ContainsUnsafeDirectives(typeDeclaration, TextSpan.FromBounds(previousBlockEnd, span.Start)))
             {
                 return false;
             }
 
-            if (SyntaxTriviaUtilities.ContainsUnbalancedConditionalDirectives(typeDeclaration, span))
+            if (ContainsUnsafeDirectives(typeDeclaration, span))
             {
                 return false;
             }
@@ -126,6 +127,18 @@ public class RH7309RegionsShouldFollowCategoryOrderCodeFixProvider : CodeFixProv
         }
 
         return true;
+    }
+
+    /// <summary>
+    /// Determines whether the span holds a directive that must not be relocated by the region swap
+    /// </summary>
+    /// <param name="typeDeclaration">Type declaration whose regions should be reordered</param>
+    /// <param name="span">Span to inspect</param>
+    /// <returns><see langword="true"/> if the span must not be moved</returns>
+    private static bool ContainsUnsafeDirectives(TypeDeclarationSyntax typeDeclaration, TextSpan span)
+    {
+        return SyntaxTriviaUtilities.ContainsUnbalancedConditionalDirectives(typeDeclaration, span)
+               || SyntaxTriviaUtilities.ContainsPositionSensitiveDirectives(typeDeclaration, span);
     }
 
     /// <summary>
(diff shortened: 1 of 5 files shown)


Reihitsu.Analyzer.CodeFixes/Rules/Organization/RH7309RegionsShouldFollowCategoryOrderCodeFixProvider.cs
```diff
@@ -82,6 +86,63 @@ private static async Task<Document> ReorderRegionsAsync(Document document, TypeD
...
+            var span = GetRegionBlockSpan(sourceText, region);
+
+            if (previousBlockEnd >= 0
+                && previousBlockEnd < span.Start
+                && SyntaxTriviaUtilities.ContainsUnbalancedConditionalDirectives(typeDeclaration, TextSpan.FromBounds(previousBlockEnd, span.Start)))
+            {
+                return false;
+            }
+
+            if (SyntaxTriviaUtilities.ContainsUnbalancedConditionalDirectives(typeDeclaration, span))
```
thoenissen: The guard only balances `#if`/`#elif`/`#else`/`#endif`, so the swap still relocates scoped non-conditional directives (`#pragma warning`, `#nullable`, `#line`) and silently changes what they cover. Verified by running the fix over this input:

```csharp
internal interface IWidget
{
    void Render();
}

internal class Widget : IWidget
{
    #region IWidget

#pragma warning disable CS0169

    public void Render()
    {
    }

    #endregion // IWidget

    #region Methods

    private int _unused;

    #endregion // Methods
}
```

The fix is offered and produces `#region Methods` with `private int _unused;` **before** the `#pragma`, so CS0169 is no longer suppressed — a build break under `TreatWarningsAsErrors`. The same shape with `#nullable enable` flips the annotation context of the moved members.

Every other reorder guard in the repo refuses on *any* directive (`OrderingDeclarationUtilities.MoveRangeContainsDirectives:256`, `AccessorOrderingUtilities.MoveRangeContainsDirectives:105`, `UsingDirectiveOrderingSafety.CanSafelyReorder:33`), and #461 asks this one to mirror `MoveRangeContainsDirectives`. Extend `CanReorderSafely` to also refuse when a swapped block or an inter-block gap contains a scoped non-conditional directive, add a `#pragma` regression test, and update the `RH7309.md` paragraph accordingly.

---

Reihitsu.Core/SyntaxTriviaUtilities.cs
```diff
@@ -194,6 +195,56 @@ public static bool IsFollowedByDirective(IEnumerable<SyntaxTrivia> trivia)
...
+    /// use this predicate to refuse the move instead
+    /// </summary>
+    /// <param name="root">Syntax node containing the span</param>
+    /// <param name="span">Span to inspect</param>
+    /// <returns><see langword="true"/> if the span contains an unbalanced conditional directive; otherwise, <see langword="false"/></returns>
+    public static bool ContainsUnbalancedConditionalDirectives(SyntaxNode root, TextSpan span)
+    {
+        if (root == null)
+        {
+            return false;
```
thoenissen: A safety predicate that fails open: a `null` root reports "no unbalanced directives", which callers read as "safe to move". Since the PR advertises this API for reuse by the remaining directive-safety fixes, return `true` here so an unusable root refuses the rewrite instead of permitting it.

---

thoenissen (commented):
The conditional-directive balance logic itself traces clean: every block/gap shape I walked (wrap-earlier, wrap-later, wrap-both, balanced-inside, balanced-between, nested `#if`, leading `#endif`, detached `#else`) resolves correctly, and moving the refusal into `RegisterCodeFixesAsync` correctly stops the no-op action from being offered. Two findings, both inline.

1. `RH7309RegionsShouldFollowCategoryOrderCodeFixProvider.cs:120` — the guard covers only conditional directives, so `#pragma warning` / `#nullable` / `#line` are still carried with the swapped block. Confirmed by execution; the defect class from #461 stays partly open.
2. `Reihitsu.Core/SyntaxTriviaUtilities.cs:212` — the new public predicate fails open on a `null` root.

---

Reihitsu.Analyzer.CodeFixes/Rules/Organization/RH7309RegionsShouldFollowCategoryOrderCodeFixProvider.cs
thoenissen: Addressed in f24c927, with the reproducing tests committed first in 6b55cd4.

`CanReorderSafely` now refuses on position-sensitive directives as well as unbalanced conditional ones. The new `SyntaxTriviaUtilities.ContainsPositionSensitiveDirectives` flags any directive in a swapped block or inter-block gap that is neither a region nor a conditional directive, so `#pragma warning`, `#nullable` and `#line` all block the swap — and unknown or malformed directive kinds fail closed rather than slipping through an enumerated list.

Tests cover your `#pragma`/CS0169 repro, the `#nullable enable` variant, and a `#pragma` in the gap between the two regions. A fourth test pins that a directive sitting ahead of *every* reordered region still gets the fix, since the swap cannot change what it covers there. `RH7309.md` now documents both withheld cases.

---

Reihitsu.Core/SyntaxTriviaUtilities.cs
thoenissen: Addressed in f24c927: `ContainsUnbalancedConditionalDirectives` now returns `true` for a `null` root, so an uninspectable span refuses the rewrite instead of permitting it. The `<returns>` documentation states that case explicitly, and a unit test pins it.

The new `ContainsPositionSensitiveDirectives` added in the same commit fails closed the same way.

---

Reihitsu.Analyzer.CodeFixes/Rules/Organization/RH7309RegionsShouldFollowCategoryOrderCodeFixProvider.cs
thoenissen: Verified fixed. Re-ran the original `#pragma warning disable CS0169` counterexample through the real analyzer diagnostic path (`ApplyCodeFixAsync`) at `c23ba9e`: no code action is registered. `ContainsPositionSensitiveDirectives` correctly excludes only region and conditional directives, so the three pre-existing balanced-conditional cases still get the fix — all 19 RH7309 tests and all 41 `SyntaxTriviaUtilitiesTests` pass.

---

Reihitsu.Core/SyntaxTriviaUtilities.cs
thoenissen: Verified fixed. Both predicates return `true` for a `null` root, the `<returns>` documentation states it, and both cases are pinned by tests.

---

PR #564, opened 2026-07-25
Remove only the flagged element in the RH8107 and RH8202 code fixes

thoenissen:
## Summary

The RH8107 and RH8202 code fixes no longer remove every line the flagged element touches. A line is still removed whole when the element is the only thing on it; as soon as another tag or documentation text shares the element's first or last line, only the element plus its adjacent horizontal whitespace is removed. Elements without an end tag are no longer offered a fix at all.

## Why

Both fixes deleted every line the flagged element touched. That silently dropped co-located documentation — a `summary` element sharing the line with the flagged `returns` element, or a `summary` end tag preceding it — and, when the element's last line carried the next tag's opening, deleted that opening too and left the comment as malformed XML.

## Linked issues

Closes #462

## Review notes

- **Trivia preservation follows from the span arithmetic.** The computed removal span is a subset of the whole-line span the old code used, in every branch: the element-owns-its-lines branch returns exactly that span, and the other three keep `start >= startLine.Start` and `end <= endLine.End`, which is strictly inside it because it stops short of the trailing line break. So the fix can only ever remove *less* than before — no trivia outside the element's own lines is reachable.
- **Malformed elements are skipped, not clamped.** When the end tag is missing, Roslyn extends the element to the end of the documentation comment, so its span runs past the trailing line break — `endLine` becomes the *next* source line and the whitespace absorption would pull the following declaration onto the `///` exterior. Clamping to the trivia does not save it: for an unterminated `returns` followed by a `param` line, the `param` element is parsed as a *child*, so a clamped span still deletes it. `CanRemoveElement` therefore skips fix registration when `EndTag.IsMissing`; the diagnostic is still reported. `main` corrupts both shapes as well, so this is a fix, not a regression.
- **Whitespace absorption is deliberately one-sided.** Content before the element means whitespace is absorbed on the left so the preceding tag keeps its separator; an exterior-only prefix means the `///` is kept and whitespace is absorbed on the right so a following tag moves up cleanly.
- **Placement.** The removal span is code-fix-only logic and now lives in `Reihitsu.Analyzer.CodeFixes/Core/DocumentationCommentCodeFixUtilities.cs`, sharing the existing `IsHorizontalWhitespace` primitive — which `RH8308`'s provider now calls too, replacing its inline copy. The exterior classification moved to `Reihitsu.Core.DocumentationCommentUtilities.IsExteriorOnly` so analyzer, code fix, and formatter agree on one policy; `DocCommentElementNormalizer` now consumes the shared `DocumentationExterior` constant. `GetLineSpanContainingSpan` is gone — it had no other callers.
- **Not done, deliberately:** the formatter's `StripDocumentationPrefix` still does not recognize the `*` continuation exterior. Routing it through the shared classifier would change formatter output for `/** … */` input, which is a behavior change rather than de-duplication and out of scope here.
- RH8304 flags the shared-line shapes this fix now preserves, so the two rules no longer conflict.

## Follow-up work

None.

### DIFF
--- a/Reihitsu.Analyzer.CodeFixes/Core/DocumentationCommentCodeFixUtilities.cs
+++ b/Reihitsu.Analyzer.CodeFixes/Core/DocumentationCommentCodeFixUtilities.cs
@@ -11,7 +11,7 @@ using Reihitsu.Formatter;
 namespace Reihitsu.Analyzer.CodeFixes.Core;
 
 /// <summary>
-/// Shared helpers for documentation-comment line-break code fixes
+/// Shared helpers for documentation-comment code fixes which rewrite line breaks or remove elements
 /// </summary>
 internal static class DocumentationCommentCodeFixUtilities
 {
--- a/Reihitsu.Analyzer.CodeFixes/Core/DocumentationCommentCodeFixUtilities.cs
+++ b/Reihitsu.Analyzer.CodeFixes/Core/DocumentationCommentCodeFixUtilities.cs
@@ -2,6 +2,7 @@
 using System.Threading.Tasks;
 
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Text;
 
 using Reihitsu.Core;
@@ -49,6 +50,87 @@ internal static class DocumentationCommentCodeFixUtilities
     }
 
     /// <summary>
+    /// Determines whether the documentation element at the specified span can be removed safely
+    /// </summary>
+    /// <param name="root">Syntax root</param>
+    /// <param name="elementSpan">Span of the documentation element</param>
+    /// <returns><see langword="true"/> if the element is well formed enough to be removed</returns>
+    /// <remarks>
+    /// An element without an end tag is left alone. Roslyn extends such an element to the end of the documentation
+    /// comment, so its span reaches past the trailing line break and can cover unrelated tags and the declaration
+    /// which follows. Removing that span would move source code into the comment or drop neighboring documentation
+    /// </remarks>
+    internal static bool CanRemoveElement(SyntaxNode root, TextSpan elementSpan)
+    {
+        var node = root.FindNode(elementSpan, findInsideTrivia: true, getInnermostNodeForTie: true);
+
+        return node is not XmlElementSyntax element
+               || element.EndTag.IsMissing == false;
+    }
+
+    /// <summary>
+    /// Gets the span which has to be removed to delete a documentation element without touching co-located content
+    /// </summary>
+    /// <param name="sourceText">Source text</param>
+    /// <param name="elementSpan">Span of the documentation element</param>
+    /// <returns>The removal span</returns>
+    /// <remarks>
+    /// Lines which carry nothing but the element are removed completely. As soon as another tag or documentation text
+    /// shares the first or the last line, only the element itself and its adjacent horizontal whitespace are removed.
+    /// The result is always contained in the lines the element occupies, so no trivia outside those lines is touched
+    /// </remarks>
+    internal static TextSpan GetElementRemovalSpan(SourceText sourceText, TextSpan elementSpan)
+    {
+        var startLine = sourceText.Lines.GetLineFromPosition(elementSpan.Start);
+        var endLine = sourceText.Lines.GetLineFromPosition(elementSpan.End);
+        var isLineStartOwnedByElement = DocumentationCommentUtilities.IsExteriorOnly(sourceText.ToString(TextSpan.FromBounds(startLine.Start, elementSpan.Start)));
+        var isLineEndOwnedByElement = string.IsNullOrWhiteSpace(sourceText.ToString(TextSpan.FromBounds(elementSpan.End, endLine.End)));
+
+        if (isLineStartOwnedByElement
+            && isLineEndOwnedByElement)
+        {
+            return TextSpan.FromBounds(startLine.Start, endLine.EndIncludingLineBreak);
+        }
+
+        var start = elementSpan.Start;
+        var end = elementSpan.End;
+
+        if (isLineEndOwnedByElement)
+        {
+            end = endLine.End;
+        }
+        else if (isLineStartOwnedByElement)
+        {
+            while (end < endLine.End
+                   && IsHorizontalWhitespace(sourceText[end]))
+            {
+                end++;
+            }
+        }
+
+        if (isLineStartOwnedByElement == false)
+        {
+            while (start > startLine.Start
+                   && IsHorizontalWhitespace(sourceText[start - 1]))
+            {
+                start--;
+            }
+        }
+
+        return TextSpan.FromBounds(start, end);
+    }
+
+    /// <summary>
+    /// Determines whether the specified character is horizontal whitespace
+    /// </summary>
+    /// <param name="value">Character to inspect</param>
+    /// <returns><see langword="true"/> if the character is a space or tab</returns>
+    internal static bool IsHorizontalWhitespace(char value)
+    {
+        return value is ' ' or '\t';
+    }
+
+    /// <summary>
     /// Gets the line break sequence for the affected line
     /// </summary>
     /// <param name="sourceText">Source text</param>
@@ -60,16 +142,6 @@ internal static class DocumentationCommentCodeFixUtilities
         return line.EndIncludingLineBreak > line.End
                    ? sourceText.ToString(TextSpan.FromBounds(line.End, line.EndIncludingLineBreak))
                    : ReihitsuFormatterHelpers.DetectEndOfLine(root);
-    }
-
-    /// <summary>
-    /// Determines whether the specified character is horizontal whitespace
-    /// </summary>
-    /// <param name="value">Character to inspect</param>
-    /// <returns><see langword="true"/> if the character is a space or tab</returns>
-    private static bool IsHorizontalWhitespace(char value)
-    {
-        return value is ' ' or '\t';
     }
 
     #endregion // Methods
--- a/Reihitsu.Analyzer.Test/Documentation/RH8107VoidReturnValueMustNotBeDocumentedAnalyzerTests.cs
+++ b/Reihitsu.Analyzer.Test/Documentation/RH8107VoidReturnValueMustNotBeDocumentedAnalyzerTests.cs
@@ -414,6 +414,47 @@ public class RH8107VoidReturnValueMustNotBeDocumentedAnalyzerTests : AnalyzerTes
     }
 
     /// <summary>
+    /// Verifies a diagnostic is reported for a returns tag nested in a remarks element
+    /// </summary>
+    /// <returns>A <see cref="Task"/> representing the asynchronous operation</returns>
+    [TestMethod]
+    public async Task VerifyDiagnosticForReturnsTagNestedInRemarks()
+    {
+        const string source = """
+                              namespace TestNamespace;
+
+                              internal class TestClass
+                              {
+                                  /// <summary>Runs the method.</summary>
+                                  /// <remarks>
+                                  /// {|#0:<returns>Nothing.</returns>|}
+                                  /// </remarks>
+                                  internal void TestMethod()
+                                  {
+                                  }
+                              }
+                              """;
+
+        const string fixedSource = """
+                                   namespace TestNamespace;
+
+                                   internal class TestClass
+                                   {
+                                       /// <summary>Runs the method.</summary>
+                                       /// <remarks>
+                                       /// </remarks>
+                                       internal void TestMethod()
+                                       {
+                                       }
+                                   }
+                                   """;
+
+        await Verify(source,
+                     fixedSource,
+                     Diagnostics(RH8107VoidReturnValueMustNotBeDocumentedAnalyzer.DiagnosticId, AnalyzerResources.RH8107MessageFormat));
+    }
+
+    /// <summary>
     /// Verifies no diagnostic is reported for a returns tag which only appears inside a code sample
     /// </summary>
     /// <returns>A <see cref="Task"/> representing the asynchronous operation</returns>
--- a/Reihitsu.Analyzer.Test/Documentation/RH8107VoidReturnValueMustNotBeDocumentedAnalyzerTests.cs
+++ b/Reihitsu.Analyzer.Test/Documentation/RH8107VoidReturnValueMustNotBeDocumentedAnalyzerTests.cs
@@ -210,6 +210,187 @@ public class RH8107VoidReturnValueMustNotBeDocumentedAnalyzerTests : AnalyzerTes
     }
 
     /// <summary>
+    /// Verifies the code fix keeps the tags on both sides of the returns tag on a shared line
+    /// </summary>
+    /// <returns>A <see cref="Task"/> representing the asynchronous operation</returns>
+    [TestMethod]
+    public async Task VerifyCodeFixKeepsSurroundingTagsSharingTheLine()
+    {
+        const string source = """
+                              namespace TestNamespace;
+
+                              internal class TestClass
+                              {
+                                  /// <summary>Runs the method.</summary>{|#0:<returns>Nothing.</returns>|}<param name="value">Value.</param>
+                                  internal void TestMethod(int value)
+                                  {
+                                  }
+                              }
+                              """;
+
+        const string fixedSource = """
+                                   namespace TestNamespace;
+
+                                   internal class TestClass
+                                   {
+                                       /// <summary>Runs the method.</summary><param name="value">Value.</param>
+                                       internal void TestMethod(int value)
+                                       {
+                                       }
+                                   }
+                                   """;
+
+        await Verify(source,
+                     fixedSource,
+                     Diagnostics(RH8107VoidReturnValueMustNotBeDocumentedAnalyzer.DiagnosticId, AnalyzerResources.RH8107MessageFormat));
+    }
+
+    /// <summary>
+    /// Verifies the code fix removes the returns tag from a delimited documentation comment
+    /// </summary>
+    /// <returns>A <see cref="Task"/> representing the asynchronous operation</returns>
+    [TestMethod]
+    public async Task VerifyCodeFixRemovesReturnsTagFromDelimitedDocumentationComment()
+    {
+        const string source = """
+                              namespace TestNamespace;
+
+                              internal class TestClass
+                              {
+                                  /**
+                                   * <summary>Runs the method.</summary>
+                                   * {|#0:<returns>Nothing.</returns>|}
+                                   */
+                                  internal void TestMethod()
+                                  {
+                                  }
+                              }
+                              """;
+
+        const string fixedSource = """
+                                   namespace TestNamespace;
+
+                                   internal class TestClass
+                                   {
+                                       /**
+                                        * <summary>Runs the method.</summary>
+                                        */
+                                       internal void TestMethod()
+                                       {
+                                       }
+                                   }
+                                   """;
+
+        await Verify(source,
+                     fixedSource,
+                     Diagnostics(RH8107VoidReturnValueMustNotBeDocumentedAnalyzer.DiagnosticId, AnalyzerResources.RH8107MessageFormat));
+    }
+
+    /// <summary>
+    /// Verifies multiple returns tags are detected and fixed together
+    /// </summary>
+    /// <returns>A <see cref="Task"/> representing the asynchronous operation</returns>
+    [TestMethod]
+    public async Task VerifyReturnsTagsAreFixedTogether()
+    {
+        const string source = """
+                              namespace TestNamespace;
+
+                              internal class TestClass
+                              {
+                                  /// <summary>Runs the method.</summary>
+                                  /// {|#0:<returns>Nothing.</returns>|}
+                                  internal void TestMethod()
+                                  {
+                                  }
+
+                                  /// <summary>Runs the other method.</summary>{|#1:<returns>Nothing.</returns>|}
+                                  internal void TestOtherMethod()
+                                  {
+                                  }
+                              }
+                              """;
+
+        const string fixedSource = """
+                                   namespace TestNamespace;
+
+                                   internal class TestClass
+                                   {
+                                       /// <summary>Runs the method.</summary>
+                                       internal void TestMethod()
+                                       {
+                                       }
+
+                                       /// <summary>Runs the other method.</summary>
+                                       internal void TestOtherMethod()
+                                       {
+                                       }
+                                   }
+                                   """;
+
+        await Verify(source,
+                     fixedSource,
+                     static config => config.NumberOfFixAllIterations = 1,
+                     Diagnostics(RH8107VoidReturnValueMustNotBeDocumentedAnalyzer.DiagnosticId, AnalyzerResources.RH8107MessageFormat, 2));
+    }
+
+    /// <summary>
+    /// Verifies no code fix is offered when the returns tag has no end tag
+    /// </summary>
+    /// <returns>A <see cref="Task"/> representing the asynchronous operation</returns>
+    /// <remarks>
+    /// Roslyn extends the element span past the trailing line break when the end tag is missing, so the following
+    /// declaration would be pulled into the documentation comment
+    /// </remarks>
+    [TestMethod]
+    public async Task VerifyNoCodeFixWhenEndTagIsMissing()
+    {
+        const string source = """
+                              namespace TestNamespace;
+
+                              internal class TestClass
+                              {
+                                  /// <summary>Runs the method.</summary>
+                                  /// <returns>Nothing.
+                                  internal void TestMethod()
+                                  {
+                                  }
+                              }
+                              """;
+
+        await Verify(source,
+                     source,
+                     Diagnostic(RH8107VoidReturnValueMustNotBeDocumentedAnalyzer.DiagnosticId).WithSpan(6, 9, 7, 1)
+                                                                                              .WithMessage(AnalyzerResources.RH8107MessageFormat));
+    }
+
+    /// <summary>
+    /// Verifies no code fix is offered when a returns tag without an end tag swallows a following tag
+    /// </summary>
+    /// <returns>A <see cref="Task"/> representing the asynchronous operation</returns>
+    [TestMethod]
+    public async Task VerifyNoCodeFixWhenEndTagIsMissingAndFollowingTagIsSwallowed()
+    {
+        const string source = """
+                              namespace TestNamespace;
+
+                              internal class TestClass
+                              {
+                                  /// <returns>Nothing.
+                                  /// <param name="value">Value.</param>
+                                  internal void TestMethod(int value)
+                                  {
+                                  }
+                              }
+                              """;
+
+        await Verify(source,
+                     source,
+                     Diagnostic(RH8107VoidReturnValueMustNotBeDocumentedAnalyzer.DiagnosticId).WithSpan(5, 9, 7, 1)
+                                                                                              .WithMessage(AnalyzerResources.RH8107MessageFormat));
+    }
+
+    /// <summary>
     /// Verifies no diagnostics are reported when documentation mode is none
     /// </summary>
     /// <returns>A <see cref="Task"/> representing the asynchronous operation</returns>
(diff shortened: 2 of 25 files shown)


Reihitsu.Analyzer/Core/DocumentationAnalysisUtilities.cs
```diff
@@ -302,17 +302,54 @@ internal static bool CanInheritDocumentation(MemberDeclarationSyntax declaration
...
+    /// <remarks>
+    /// Lines which carry nothing but the element are removed completely. As soon as another tag or documentation text
+    /// shares the first or the last line, only the element itself and its adjacent whitespace are removed.
+    /// </remarks>
+    internal static TextSpan GetDocumentationElementRemovalSpan(SourceText text, TextSpan span)
     {
         var startLine = text.Lines.GetLineFromPosition(span.Start);
         var endLine = text.Lines.GetLineFromPosition(span.End);
+        var isLineStartOwnedByElement = IsDocumentationExteriorOnly(text.ToString(TextSpan.FromBounds(startLine.Start, span.Start)));
+        var isLineEndOwnedByElement = string.IsNullOrWhiteSpace(text.ToString(TextSpan.FromBounds(span.End, endLine.End)));
```
thoenissen: The span math assumes the element ends on its own last line. When the end tag is missing, Roslyn's element span runs past the trailing line break, so `endLine` is the **next** source line: `isLineEndOwnedByElement` becomes false and the loop on line 336 absorbs that line's indentation, pulling non-documentation content onto the `///` exterior.

```cs
/// <summary>Runs.</summary>
/// <returns>Nothing.
internal void Save()
{
}
```

becomes

```cs
/// <summary>Runs.</summary>
/// internal void Save()
{
}
```

— CS1519 and CS1022; the declaration is swallowed into the comment. The same path destroys directives: with `#if DEBUG` inside the element, `#if DEBUG` becomes `/// #if DEBUG` and the `#endif` is orphaned (CS1028). Clamp the removal span to the documentation comment trivia, or refuse the fix when the element's end tag is missing. `main` corrupts both inputs too (it deletes the following line outright), so this is not a regression — but it is the defect class this PR closes and it stays open.

---

Reihitsu.Analyzer/Core/DocumentationAnalysisUtilities.cs
```diff
@@ -417,6 +454,32 @@ internal static bool MatchesAccessibilityGroup(EnumMemberDeclarationSyntax decla
                && MatchesAccessibilityGroup(enumDeclaration, accessibilityGroup);
     }
 
+    /// <summary>
+    /// Determines whether the text consists of whitespace and documentation comment exterior markers only
+    /// </summary>
+    /// <param name="text">Text</param>
+    /// <returns><see langword="true"/> if the text carries no documentation content</returns>
+    private static bool IsDocumentationExteriorOnly(string text)
```
thoenissen: This is a third private encoding of the documentation-exterior marker: `Reihitsu.Core.DocumentationCommentUtilities` owns `DocumentationExterior = "///"`, and `Reihitsu.Formatter`'s `DocCommentElementNormalizer.StripDocumentationPrefix` hard-codes `"///"` again. The three disagree — only this copy recognizes `/** … * … */` exteriors. Move the exterior classification into `Reihitsu.Core.DocumentationCommentUtilities` so analyzer, code fix, and formatter share one policy.

---

Reihitsu.Analyzer/Core/DocumentationAnalysisUtilities.cs
```diff
@@ -302,17 +302,54 @@ internal static bool CanInheritDocumentation(MemberDeclarationSyntax declaration
...
-    /// <param name="span">Source span</param>
-    /// <returns>The line span</returns>
-    internal static TextSpan GetLineSpanContainingSpan(SourceText text, TextSpan span)
+    /// <param name="span">Span of the documentation element</param>
+    /// <returns>The removal span</returns>
+    /// <remarks>
+    /// Lines which carry nothing but the element are removed completely. As soon as another tag or documentation text
+    /// shares the first or the last line, only the element itself and its adjacent whitespace are removed.
+    /// </remarks>
+    internal static TextSpan GetDocumentationElementRemovalSpan(SourceText text, TextSpan span)
```
thoenissen: This is code-fix-only logic — a removal span — placed in the analyzer assembly; no analyzer calls it. `Reihitsu.Analyzer.CodeFixes/Core/DocumentationCommentCodeFixUtilities.cs` is the existing home for documentation-comment fix helpers and already contains the same "absorb adjacent whitespace" loop (`IsHorizontalWhitespace`), which `RH8308NoContentShouldAppearAfterClosingXmlTagsCodeFixProvider` copies inline as well; this adds a third copy, using `char.IsWhiteSpace` instead. Move the helper there and share one whitespace-absorption primitive.

---

Reihitsu.Analyzer.Test/Documentation/RH8107VoidReturnValueMustNotBeDocumentedAnalyzerTests.cs
```diff
@@ -54,6 +54,161 @@ await Verify(source,
...
+                     fixedSource,
+                     Diagnostics(RH8107VoidReturnValueMustNotBeDocumentedAnalyzer.DiagnosticId, AnalyzerResources.RH8107MessageFormat));
+    }
+
+    /// <summary>
+    /// Verifies the code fix removes every line of a multi line returns tag which owns its lines
+    /// </summary>
+    /// <returns>A <see cref="Task"/> representing the asynchronous operation</returns>
+    [TestMethod]
+    public async Task VerifyCodeFixRemovesMultiLineReturnsTagCompletely()
```
thoenissen, line 178: Three paths this change introduces have no test. FixAll: both providers expose `WellKnownFixAllProviders.BatchFixer` and several void members with `<returns>` co-occur trivially — the repo's convention is `config.NumberOfFixAllIterations` (see `RH8301DocumentationLinesMustBeginWithSingleSpaceAnalyzerTests`). A `/** … */` documentation comment: it is the only reason `IsDocumentationExteriorOnly` accepts an all-`*` marker and rejects `/**`, and the PR notes call that behavior deliberate. Content on both sides of the element on one line (`/// <summary>X</summary><returns>y</returns><remarks>z</remarks>`): the fourth branch of the new helper. The same three gaps apply to `RH8202ValueTagMustNotBeUsedAnalyzerTests`.

---

thoenissen (commented):
Reviewed the RH8107/RH8202 removal-span change against issue #462 and against every shape the new helper branches on. Verified by targeted execution on this branch (048c40b) and on the base (b7bb0be): the three shapes the issue requires are fixed, and the `/** … */`, both-sides-shared and CRLF shapes behave correctly. Two confirmed defects remain — one inline, one below.

1. `Reihitsu.Analyzer/Core/DirectDocumentationSyntaxChecker.cs` — `GetDirectTags` falls back to `DescendantNodes()` when the documentation comment has no top-level match, so RH8107 and RH8202 flag a `<returns>`/`<value>` tag that appears inside a `<code>` or `<example>` sample and the fix deletes it. Confirmed on this branch:

```cs
/// <summary>
/// Saves the customer.
/// <code>
/// <returns>Sample tag inside a code block.</returns>
/// </code>
/// </summary>
internal void Save()
{
}
```

The `<returns>` line inside `<code>` is removed. This is the same "the fix must not remove content it does not own" class as #462 and it is not closed here. The fallback also contradicts the class name and the `GetDirectTags` doc comment ("first **direct** XML node"), and it affects every `GetDirectTags`/`HasDirectTag` caller, not only these two rules.

---

Reihitsu.Analyzer/Core/DocumentationAnalysisUtilities.cs
thoenissen: Addressed in 606da99. Confirmed both shapes against Roslyn before fixing: with the end tag missing the element span is `<returns>Nothing.\n` — it ends at the *start* of the next line, so `endLine` is the declaration's line. Two regression tests now pin this (`VerifyNoCodeFixWhenEndTagIsMissing`, `VerifyNoCodeFixWhenEndTagIsMissingAndFollowingTagIsSwallowed`, plus the RH8202 pair); before the fix they produced `/// internal void TestMethod()` exactly as you described.

I took the second option — refuse the fix — rather than clamping. Clamping to the trivia does not cover the second shape: for

```cs
/// <returns>Nothing.
/// <param name="value">Value.</param>
internal void TestMethod(int value)
```

Roslyn nests the `param` element inside the unterminated `returns` element, so the span is `<returns>Nothing.\n    /// <param name="value">Value.</param>\n` — entirely inside the documentation comment. A clamp accepts it and still deletes the `param` line. `DocumentationCommentCodeFixUtilities.CanRemoveElement` now skips registration when `EndTag.IsMissing`, so the diagnostic is still reported but no corrupting fix is offered.

---

Reihitsu.Analyzer/Core/DocumentationAnalysisUtilities.cs
thoenissen: Partially addressed in 606da99. The classification moved to `Reihitsu.Core.DocumentationCommentUtilities.IsExteriorOnly`, `DocumentationExterior` is now public, and `DocCommentElementNormalizer.StripDocumentationPrefix` consumes that constant instead of its own `"///"` literal.

What I did **not** do: make the formatter's `StripDocumentationPrefix` accept the `*` continuation exterior. That is a behavior change for `/** … */` input, not a de-duplication — today the formatter falls through to `rawLine.Trim()` for those lines, and routing them through the shared policy would change formatter output. It is out of scope for a code-fix span bug, so the three call sites now share one *constant* and one *classifier*, but the formatter's `/** … */` handling is unchanged. Flagging it here rather than silently widening the fix.

---

Reihitsu.Analyzer/Core/DocumentationAnalysisUtilities.cs
thoenissen: Addressed in 606da99. The helper moved to `Reihitsu.Analyzer.CodeFixes/Core/DocumentationCommentCodeFixUtilities.cs` as `GetElementRemovalSpan`, and both loops now call the existing `IsHorizontalWhitespace` instead of `char.IsWhiteSpace` — which is also the more correct primitive here, since the spans are already clamped to line bounds and a line break must never be absorbed. `RH8308NoContentShouldAppearAfterClosingXmlTagsCodeFixProvider` now calls the same primitive rather than its inline `' '`/`'\t'` comparison, so the third copy is gone. `DocumentationAnalysisUtilities` lost both `GetLineSpanContainingSpan` and the private classifier; the analyzer assembly no longer carries any removal-span logic.

---

Reihitsu.Analyzer.Test/Documentation/RH8107VoidReturnValueMustNotBeDocumentedAnalyzerTests.cs
thoenissen: Addressed in 606da99 — all three gaps covered for both rules:

- **FixAll**: `VerifyReturnsTagsAreFixedTogether` / `VerifyValueTagsAreFixedTogether`, using `config.NumberOfFixAllIterations = 1` per the `RH8301` convention. Each mixes an element-owns-its-line occurrence with a shared-line one so the batch exercises two different branches.
- **`/** … */`**: `VerifyCodeFixRemovesReturnsTagFromDelimitedDocumentationComment` / `...ValueTag...`. Confirmed this path is live — `GetDocumentationComment` selects `DocumentationCommentTriviaSyntax`, which covers both single-line and delimited trivia, so the all-`*` branch is reachable and now pinned.
- **Content on both sides**: `VerifyCodeFixKeepsSurroundingTagsSharingTheLine` for both rules, covering the fourth branch (both flags false).

Analyzer suite is at 2140 tests, all green.

---

Reihitsu.Analyzer/Core/DocumentationAnalysisUtilities.cs
thoenissen: Verified on 182f7aa: both shapes now register no code action — the unterminated element and the `#if DEBUG` inside the element. `<returns/>` is still fixed, so the guard does not over-refuse. Refusing rather than clamping is the right call; your nesting argument holds. Resolved.

---

Reihitsu.Analyzer/Core/DocumentationAnalysisUtilities.cs
thoenissen: Resolved. The classifier is now `Reihitsu.Core.DocumentationCommentUtilities.IsExteriorOnly`, the analyzer's private copy is gone, and the formatter's `"///"` literal is replaced by the shared constant — the duplication this finding named no longer exists. Your scope call on `StripDocumentationPrefix` is accepted. Tracing it did surface a separate confirmed formatter defect for `/** … */` comments, reported in the new review summary rather than here.

---

Reihitsu.Analyzer/Core/DocumentationAnalysisUtilities.cs
thoenissen: Resolved. `GetElementRemovalSpan` and `CanRemoveElement` live in the code fix assembly, both loops call the shared `IsHorizontalWhitespace`, `RH8308` uses it instead of its inline comparison, and the analyzer assembly no longer carries removal-span logic. Two leftovers noted in the new review summary.

---

Reihitsu.Analyzer.Test/Documentation/RH8107VoidReturnValueMustNotBeDocumentedAnalyzerTests.cs
thoenissen: Resolved. All three gaps are covered for both rules, and the two malformed-element regression tests pin the guard. Verified independently that the `/** … */` path is live and that the delimited and both-sides-shared shapes produce the asserted output.

---

Reihitsu.Analyzer.CodeFixes/Core/DocumentationCommentCodeFixUtilities.cs
```diff
@@ -49,28 +50,99 @@ internal static async Task<Document> MoveContentToNextDocumentationLineAsync(Doc
...
+    /// </summary>
+    /// <param name="root">Syntax root</param>
+    /// <param name="elementSpan">Span of the documentation element</param>
+    /// <returns><see langword="true"/> if the element is well formed enough to be removed</returns>
+    /// <remarks>
+    /// An element without an end tag is left alone. Roslyn extends such an element to the end of the documentation
+    /// comment, so its span reaches past the trailing line break and can cover unrelated tags and the declaration
+    /// which follows. Removing that span would move source code into the comment or drop neighboring documentation
+    /// </remarks>
+    internal static bool CanRemoveElement(SyntaxNode root, TextSpan elementSpan)
```
thoenissen, line 63: The class summary on line 14 still reads "Shared helpers for documentation-comment **line-break** code fixes", but the class now also owns the removal span and the malformed-element guard. Widen the summary to cover element removal.

---

thoenissen (commented):
Re-review of 182f7aa. All four inline findings from the previous pass are verified fixed and their threads are resolved: the guard refuses the unterminated element *and* the `#if`-inside-element shape (both now register no code action), `<returns/>` is still fixed so it does not over-refuse, the removal span and whitespace primitive live in the code fix assembly, the exterior classifier is shared through `Reihitsu.Core`, and all three test gaps are closed for both rules. One finding from the previous summary is still open, and tracing your `/** … */` scope decision surfaced a confirmed formatter defect.

1. **Still open** — `Reihitsu.Analyzer/Core/DirectDocumentationSyntaxChecker.cs`: `GetDirectTags` falls back to `DescendantNodes()` when the documentation comment has no top-level match, so RH8107 and RH8202 flag a `<returns>`/`<value>` tag inside a `<code>` or `<example>` sample and the fix deletes it. Reconfirmed on 182f7aa — `CanRemoveElement` does not catch it, because the nested element is well formed:

```cs
/// <summary>
/// Saves the customer.
/// <code>
/// <returns>Sample tag inside a code block.</returns>
/// </code>
/// </summary>
internal void Save()
{
}
```

The `<returns>` line inside `<code>` is removed. Same defect class as #462; the fallback also contradicts the class name and the `GetDirectTags` doc comment ("first **direct** XML node"), and it affects every `GetDirectTags`/`HasDirectTag` caller.

2. **New** — the formatter deletes the line break after every `/** … */` documentation comment, joining the declaration onto the closing delimiter. `dotnet run --project Reihitsu.Cli -- <file>` over

```cs
internal class TestClass
{
    /** <summary>Runs.</summary> */
    internal void Save()
    {
    }
}
```

produces `    /** <summary>Runs.</summary> */    internal void Save()`. It reproduces for the expanded form as well (`*` continuation lines, no element needing normalization), is stable on a second pass, and does not occur for a plain `/* … */` block comment — so it is specific to `MultiLineDocumentationCommentTrivia`. Identical on b7bb0be, so it is pre-existing rather than caused by this PR, and it is not `DocCommentElementNormalizer`: shapes that need no normalization are corrupted too.

3. **New** — `Reihitsu.Analyzer/Core/DocumentationAnalysisUtilities.cs` line 8: `using Microsoft.CodeAnalysis.Text;` is now dead. Removing `GetLineSpanContainingSpan` took the last `SourceText`/`TextSpan` reference in the file with it.

---

Reihitsu.Analyzer.CodeFixes/Core/DocumentationCommentCodeFixUtilities.cs
thoenissen: Addressed in 4755311. Summary now reads "Shared helpers for documentation-comment code fixes which rewrite line breaks or remove elements".

---

thoenissen: Verified on d1c3fb2. Resolved.

---

Reihitsu.Analyzer/Core/DirectDocumentationSyntaxChecker.cs
```diff
@@ -38,6 +39,29 @@ internal static XmlNodeSyntax GetFirstDirectTag(DocumentationCommentTriviaSyntax
...
+    /// </summary>
+    /// <param name="documentationComment">Documentation comment</param>
+    /// <param name="tagName">Tag name</param>
+    /// <returns>The matching node, when present</returns>
+    /// <remarks>
+    /// Unlike <see cref="GetFirstDirectTag"/> this never falls back to nested nodes, so a tag written inside a
+    /// <c>&lt;code&gt;</c> or <c>&lt;example&gt;</c> sample is never returned. Rules which remove the tag they
+    /// find must use this overload, otherwise the fix deletes sample text the documentation only illustrates
+    /// </remarks>
+    internal static XmlNodeSyntax GetFirstTopLevelTag(DocumentationCommentTriviaSyntax documentationComment, string tagName)
```
thoenissen: This excludes every nested tag, not only tags inside samples, so RH8107 and RH8202 stopped reporting real violations. Verified on d1c3fb2 versus b7bb0be — both of these were flagged on the base and are silent now:

```cs
/// <summary>Gets the value.</summary>
/// <remarks>
/// <value>The current value.</value>
/// </remarks>
internal int Value { get; }
```

```cs
/// <summary>Gets the <value>current</value> value.</summary>
internal int Value { get; }
```

Neither is a `<code>`/`<example>` sample, and RH8202's own doc says the `<value>` tag must not be used. Keep the descendant search and exclude only elements whose ancestors include `<code>` or `<example>`; alternatively keep detection and gate the *fix* on the tag being top level, which is where the deletion hazard actually lives.

---

Reihitsu.Analyzer/Core/DirectDocumentationSyntaxChecker.cs
```diff
@@ -38,6 +39,29 @@ internal static XmlNodeSyntax GetFirstDirectTag(DocumentationCommentTriviaSyntax
...
+    {
+        if (documentationComment == null)
+        {
+            return null;
+        }
+
+        return documentationComment.Content
+                                   .Where(obj => obj is XmlElementSyntax or XmlEmptyElementSyntax)
```
thoenissen: This query is a copy of the `directNodes` block in `GetDirectTags` below it. Extract the top-level query once and have `GetDirectTags` call it before falling back. The naming also inverts now: `GetFirstDirectTag`/`GetDirectTags` are the members that search descendants, while `GetFirstTopLevelTag` is the one that is actually direct — rename the fallback members to say they search descendants.

---

thoenissen (commented):
Re-review of d1c3fb2. The dead using and the class summary are fixed and that thread is resolved. The sample-deletion path is closed for RH8107/RH8202 — verified `<returns>` inside `<code>` reports nothing on d1c3fb2 where b7bb0be reported RH8107 at 7,8. But the narrowing overshot, and two findings carry over.

1. **Still open** — the shared `GetDirectTags` fallback. Your scope argument for not touching the other 11 rules is accepted, but the finding is not closed, so recording where it stands: RH8203 still fires on an `<inheritdoc/>` written inside a `<code>` sample, identically on d1c3fb2 and b7bb0be (RH8203 at 5,4 on both):

```cs
/// <summary>
/// Standalone type.
/// <code>
/// <inheritdoc/>
/// </code>
/// </summary>
internal class TestClass
{
}
```

The destructive half is fixed; the false-positive half remains for the rules that only report.

2. **Still open** — the formatter drops the line break after every `/** … */` documentation comment. Re-verified on d1c3fb2: `    /** <summary>Runs.</summary> */    internal void Save()`. Pre-existing and correctly scoped out of this PR; recorded so it is not lost.

---

Reihitsu.Analyzer/Core/DirectDocumentationSyntaxChecker.cs
thoenissen: You are right, and the over-suppression was mine. Fixed in ae744af by taking your first option.

Both of your shapes now reproduce as failing tests before the change — `VerifyDiagnosticForValueTagNestedInRemarks`, `VerifyDiagnosticForValueTagNestedInSummaryText`, plus `VerifyDiagnosticForReturnsTagNestedInRemarks` for RH8107. `GetFirstTopLevelTag` is replaced by `GetFirstDocumentingTag`, which keeps the descendant search and filters only nodes with a `<code>` or `<example>` ancestor:

```cs
return GetTagsIncludingNested(documentationComment, tagName).FirstOrDefault(obj => IsInsideSample(obj) == false);
```

So prose-nested tags are violations again, sample content stays untouched, and the convergence guards from the previous round still hold — after the top-level tag is removed the remaining sample match is filtered, so no second diagnostic appears.

I did not take the gate-the-fix-only option: it would leave RH8107/RH8202 reporting on `<code>` samples, and a tag illustrating usage in a sample is not a violation of either rule to begin with — suppressing the fix but keeping the diagnostic would just move the false positive rather than remove it.

One deliberate detail: the inline case removes the leading space with the element, so `Gets the <value>current</value> value.` becomes `Gets the value.` rather than leaving a double space.

---

Reihitsu.Analyzer/Core/DirectDocumentationSyntaxChecker.cs
thoenissen: Both parts addressed in ae744af.

**Duplication** — gone by construction. The replacement `GetFirstDocumentingTag` delegates to the existing query and filters its result, so there is no second copy of the top-level block to extract.

**Naming** — renamed across all 16 call sites in 13 files:

- `GetDirectTags` → `GetTagsIncludingNested`
- `GetFirstDirectTag` → `GetFirstTagIncludingNested`
- `HasDirectTag` → `HasTagIncludingNested`

Pure rename, no behavior change; build is clean at 0 warnings and the full suite is green at 2148 analyzer tests.

I left the `DirectDocumentationSyntaxChecker` class name alone. "Direct" there means syntax-only inspection without semantic expansion — which is what the class summary says and what it still does — rather than a claim about tag nesting, so it is not the same inversion the method names carried.

---

Reihitsu.Analyzer/Core/DirectDocumentationSyntaxChecker.cs
thoenissen: Verified on bc6ea55: both shapes report again — RH8202 at 7,8 for the `<remarks>` nesting and at 5,26 for the inline `<summary>` case — while `<returns>` in `<code>` and `<value>` in `<example>` stay silent. The fixes remove only the flagged element and leave the surrounding prose intact, and both the top-level-beside-sample and sample-before-prose shapes re-analyze clean after one pass. Resolved.

---

Reihitsu.Analyzer/Core/DirectDocumentationSyntaxChecker.cs
thoenissen: Resolved. `GetFirstDocumentingTag` delegates to the existing query, so the copied block is gone, and the three nested-search members are renamed consistently across all call sites. Agreed on keeping the class name. One leftover from the rename is in the new review.

---

Reihitsu.Analyzer/Core/DirectDocumentationSyntaxChecker.cs
```diff
@@ -44,7 +77,7 @@ internal static XmlNodeSyntax GetFirstDirectTag(DocumentationCommentTriviaSyntax
     /// <param name="documentationComment">Documentation comment</param>
     /// <param name="tagName">Tag name</param>
     /// <returns>Matching nodes</returns>
-    internal static ImmutableArray<XmlNodeSyntax> GetDirectTags(DocumentationCommentTriviaSyntax documentationComment, string tagName)
+    internal static ImmutableArray<XmlNodeSyntax> GetTagsIncludingNested(DocumentationCommentTriviaSyntax documentationComment, string tagName)
```
thoenissen, line 82: The rename did not reach the doc comments, so all three still claim "direct" — the exact inversion the rename removed from the names. Line 46 "Gets the first direct XML node", line 75 "Gets all direct XML nodes", line 105 "Determines whether the documentation contains a direct tag". Reword them to say the search includes nested nodes.

---

thoenissen (commented):
Re-review of bc6ea55. Both inline findings from the previous pass are verified fixed and resolved: prose-nested tags report again (RH8202 at 7,8 for `<remarks>`, at 5,26 for inline `<summary>`), sample content stays suppressed for both `<code>` and `<example>`, the fixes remove only the flagged element, and the top-level-beside-sample and sample-before-prose shapes both re-analyze clean after one pass. One finding carries over, one small leftover is inline.

1. **Still open** — the nested-search fallback for the rules that only report. RH8203 fires on an `<inheritdoc/>` written inside a `<code>` sample, identically on bc6ea55 and b7bb0be (RH8203 at 5,4 on both):

```cs
/// <summary>
/// Standalone type.
/// <code>
/// <inheritdoc/>
/// </code>
/// </summary>
internal class TestClass
{
}
```

`GetFirstDocumentingTag` now exists and would close this for any rule that adopts it; the 11 report-only rules still call `GetFirstTagIncludingNested`/`GetTagsIncludingNested`. Recorded, not demanded in this PR.

2. **Still open** — the formatter drops the line break after every `/** … */` documentation comment (`    /** <summary>Runs.</summary> */    internal void Save()`). No formatter file changed since d1c3fb2, where it was last reproduced. Pre-existing and correctly scoped out; recorded so it is not lost.

---

Reihitsu.Analyzer/Core/DirectDocumentationSyntaxChecker.cs
thoenissen: Addressed in b166aa7. All three summaries now describe the fallback instead of claiming "direct":

- `GetFirstTagIncludingNested` — "Gets the first XML node with the specified tag name, falling back to nested nodes when the documentation comment has no top-level match"
- `GetTagsIncludingNested` — "Gets the XML nodes with the specified tag name. Top-level nodes win; when the documentation comment has no top-level match the search falls back to nested nodes, including content inside samples"
- `HasTagIncludingNested` — "Determines whether the documentation contains the tag, at top level or nested"

Two uses of "direct" are left on purpose: the class name and its summary, in the syntax-only-inspection sense we agreed to keep, and the `directNodes` local inside `GetTagsIncludingNested`, which genuinely holds the direct children before the fallback runs.

---

thoenissen: Verified on 65f1321 — all three summaries now describe the fallback, and the two retained uses of "direct" are correct. Resolved.

---

thoenissen (commented):
Re-review of 65f1321. No new findings. The doc-comment wording is fixed and resolved, the RH8202 `<example>` case is now mirrored, and no stale member name survives anywhere in the tree. Re-checked the behavior the last two rounds established, this time against the advanced base (5353217, which merged #563 and #565): `<value>` in `<example>` is silent, `<value>` in `<remarks>` reports at 7,8, and the build is clean after the merge, so main's RH8201 change and the rename did not collide.

The two carried-over items — the nested-search fallback for the 11 report-only rules (RH8203 still at 5,4 on an `<inheritdoc/>` inside `<code>`) and the formatter dropping the line break after `/** … */` comments — were re-verified as unchanged on 65f1321. #565 touched `BlankLineCollapser`/`BlankLineTokenCleanupRewriter`, so the formatter one was worth re-running; it still reproduces. Both remain recorded here rather than restated, and both stay correctly scoped out of this PR.

---

PR #565, opened 2026-07-25
Replace multi-line documentation comments in the RH8201 code fix

thoenissen:
## Summary

`RH8201InheritdocShouldBeUsedCodeFixProvider.ReplaceDocumentation` now matches multi-line (`/** ... */`) documentation comments in addition to single-line ones, so a `/** ... */`-documented override is rewritten to `/// <inheritdoc/>` instead of the code action returning the document unchanged. Because the synthesized `<inheritdoc/>` trivia carries its own line break while a `/** ... */` comment does not, the line break that terminated the replaced comment is dropped so no blank line is introduced before the member.

The predicate that decides what counts as a documentation comment is now `SyntaxTriviaUtilities.IsDocumentationCommentTrivia` in `Reihitsu.Core`, shared by the analyzer, this code fix, and the formatter's blank-line phase.

## Why

The analyzer reports both documentation-comment kinds, but the code fix only rewrote the single-line kind. On a `/** ... */`-documented override the lightbulb action was offered and applying it did nothing — a registered no-op code action. That drift was possible because the analyzer and its fix each carried a private copy of the same two-kind predicate.

## Linked issues

Closes #463

## Review notes

- The analyzer's reporting is unchanged; this is a code-fix behavior change, so no diagnostic was rescoped or removed.
- Analyzer/fix parity is now structural: `SyntaxTriviaUtilities.IsDocumentationCommentTrivia` is the single definition, `IsCommentTrivia` delegates to it, and the three surviving copies in `Reihitsu.Formatter` (`BlankLineTokenCleanupRewriter` ×2, `BlankLineCollapser`) were replaced with it. Those formatter substitutions are behavior-preserving — each tested the same two syntax kinds.
- Whitespace between the replaced comment and its line break is skipped along with the line break. Restoring it is unnecessary: the code action's post-processing re-indents the affected lines, so the `/** doc */ // trailing` and `/** doc */ public override ...` shapes come out identical either way.
- Only the first documentation comment in the member's leading trivia is replaced, matching the analyzer's `FirstOrDefault` selection and the pre-existing single-line behavior. Where a second documentation comment survives, the compiler concatenates both into one XML `<member>` entry and reports no warning, so no documentation is lost.
- The line break is dropped in the code fix rather than delegated to the formatter, because the formatter's blank-line cleanup has a known defect around `/** */` comments (#526).
- Coverage: nine RH8201 analyzer tests (multi-line block form, single-line `/** */` form, property member, preserved leading comment, second documentation comment left intact, CRLF preservation, trailing comment in the trivia gap, preprocessor directive in the trivia gap, and Fix All over two documented overrides) plus four `SyntaxTriviaUtilities` tests for the extracted predicate.

## Follow-up work

- RH8201 registration still misses `EventFieldDeclaration`, so documented field-like override events are not checked. That minor is tracked separately in #467 and is intentionally left out of this fix-scoped change.

### DIFF
--- a/Reihitsu.Analyzer.CodeFixes/Rules/Documentation/RH8201InheritdocShouldBeUsedCodeFixProvider.cs
+++ b/Reihitsu.Analyzer.CodeFixes/Rules/Documentation/RH8201InheritdocShouldBeUsedCodeFixProvider.cs
@@ -11,6 +11,7 @@ using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 
 using Reihitsu.Analyzer.Rules.Documentation;
+using Reihitsu.Core;
 using Reihitsu.Formatter;
 
 namespace Reihitsu.Analyzer.CodeFixes.Rules.Documentation;
@@ -53,18 +54,15 @@ public class RH8201InheritdocShouldBeUsedCodeFixProvider : CodeFixProvider
     {
         var replaced = false;
         var isLineBreakPending = false;
-        var pendingTrivia = new List<SyntaxTrivia>();
 
         foreach (var trivia in triviaList)
         {
             if (isLineBreakPending)
             {
-                // Whitespace between the replaced comment and its line break is buffered so it can be restored
-                // when no line break follows after all
+                // Whitespace between the replaced comment and its line break is dropped along with the line break.
+                // Where no line break follows, the code action's post-processing re-indents the affected lines
                 if (trivia.IsKind(SyntaxKind.WhitespaceTrivia))
                 {
-                    pendingTrivia.Add(trivia);
-
                     continue;
                 }
 
@@ -72,24 +70,14 @@ public class RH8201InheritdocShouldBeUsedCodeFixProvider : CodeFixProvider
 
                 if (trivia.IsKind(SyntaxKind.EndOfLineTrivia))
                 {
-                    pendingTrivia.Clear();
-
                     continue;
                 }
-
-                foreach (var buffered in pendingTrivia)
-                {
-                    yield return buffered;
-                }
-
-                pendingTrivia.Clear();
             }
 
             // Only the flagged (first) documentation comment is replaced. Replacing every documentation comment would
             // emit multiple <inheritdoc/> lines when a member carries more than one documentation comment
             if (replaced == false
-                && (trivia.IsKind(SyntaxKind.SingleLineDocumentationCommentTrivia)
-                    || trivia.IsKind(SyntaxKind.MultiLineDocumentationCommentTrivia)))
+                && SyntaxTriviaUtilities.IsDocumentationCommentTrivia(trivia))
             {
                 replaced = true;
 
@@ -104,11 +92,6 @@ public class RH8201InheritdocShouldBeUsedCodeFixProvider : CodeFixProvider
             }
 
             yield return trivia;
-        }
-
-        foreach (var buffered in pendingTrivia)
-        {
-            yield return buffered;
         }
     }
 
--- a/Reihitsu.Analyzer.CodeFixes/Rules/Documentation/RH8201InheritdocShouldBeUsedCodeFixProvider.cs
+++ b/Reihitsu.Analyzer.CodeFixes/Rules/Documentation/RH8201InheritdocShouldBeUsedCodeFixProvider.cs
@@ -43,7 +43,8 @@ public class RH8201InheritdocShouldBeUsedCodeFixProvider : CodeFixProvider
     }
 
     /// <summary>
-    /// Replacing the first <see cref="SyntaxKind.SingleLineDocumentationCommentTrivia"/> with a &amp;lt;inheritdoc/&amp;gt; trivia
+    /// Replacing the first <see cref="SyntaxKind.SingleLineDocumentationCommentTrivia"/> or
+    /// <see cref="SyntaxKind.MultiLineDocumentationCommentTrivia"/> with a &amp;lt;inheritdoc/&amp;gt; trivia
     /// </summary>
     /// <param name="triviaList">List of trivia elements</param>
     /// <param name="endOfLine">End-of-line sequence to use for the trailing line break</param>
@@ -51,15 +52,51 @@ public class RH8201InheritdocShouldBeUsedCodeFixProvider : CodeFixProvider
     private IEnumerable<SyntaxTrivia> ReplaceDocumentation(SyntaxTriviaList triviaList, string endOfLine)
     {
         var replaced = false;
+        var isLineBreakPending = false;
+        var pendingTrivia = new List<SyntaxTrivia>();
 
         foreach (var trivia in triviaList)
         {
+            if (isLineBreakPending)
+            {
+                // Whitespace between the replaced comment and its line break is buffered so it can be restored
+                // when no line break follows after all
+                if (trivia.IsKind(SyntaxKind.WhitespaceTrivia))
+                {
+                    pendingTrivia.Add(trivia);
+
+                    continue;
+                }
+
+                isLineBreakPending = false;
+
+                if (trivia.IsKind(SyntaxKind.EndOfLineTrivia))
+                {
+                    pendingTrivia.Clear();
+
+                    continue;
+                }
+
+                foreach (var buffered in pendingTrivia)
+                {
+                    yield return buffered;
+                }
+
+                pendingTrivia.Clear();
+            }
+
             // Only the flagged (first) documentation comment is replaced. Replacing every documentation comment would
             // emit multiple <inheritdoc/> lines when a member carries more than one documentation comment
             if (replaced == false
-                && trivia.IsKind(SyntaxKind.SingleLineDocumentationCommentTrivia))
+                && (trivia.IsKind(SyntaxKind.SingleLineDocumentationCommentTrivia)
+                    || trivia.IsKind(SyntaxKind.MultiLineDocumentationCommentTrivia)))
             {
                 replaced = true;
+
+                // A /** ... */ comment does not carry its own line break, while the synthesized <inheritdoc/> trivia
+                // does. The line break that terminated the replaced comment is therefore dropped so the fix does not
+                // introduce a blank line before the member
+                isLineBreakPending = trivia.IsKind(SyntaxKind.MultiLineDocumentationCommentTrivia);
 
                 yield return CreateInheritdocTrivia(endOfLine);
 
@@ -68,11 +105,17 @@ public class RH8201InheritdocShouldBeUsedCodeFixProvider : CodeFixProvider
 
             yield return trivia;
         }
+
+        foreach (var buffered in pendingTrivia)
+        {
+            yield return buffered;
+        }
     }
 
     /// <summary>
-    /// Applies the code fix by replacing <see cref="SyntaxKind.SingleLineDocumentationCommentTrivia"/>
-    /// with a &lt;inheritdoc/&gt; trivia in the leading trivia of the specified <see cref="MemberDeclarationSyntax"/>
+    /// Applies the code fix by replacing <see cref="SyntaxKind.SingleLineDocumentationCommentTrivia"/> or
+    /// <see cref="SyntaxKind.MultiLineDocumentationCommentTrivia"/> with a &lt;inheritdoc/&gt; trivia in the
+    /// leading trivia of the specified <see cref="MemberDeclarationSyntax"/>
     /// </summary>
     /// <param name="document">The <see cref="Document"/> to apply the fix to</param>
     /// <param name="memberDeclaration">The <see cref="MemberDeclarationSyntax"/> to fix</param>
--- a/Reihitsu.Analyzer.Test/Documentation/RH8201InheritdocShouldBeUsedAnalyzerTests.cs
+++ b/Reihitsu.Analyzer.Test/Documentation/RH8201InheritdocShouldBeUsedAnalyzerTests.cs
@@ -604,7 +604,13 @@ public class RH8201InheritdocShouldBeUsedAnalyzerTests : AnalyzerTestsBase<RH820
 
     /// <summary>
     /// Verifies that only the flagged multi-line (/** */) documentation comment is replaced when a member
-    /// carries a second documentation comment (issue #463)
+    /// carries a second documentation comment (issue #463).
+    /// The surviving second comment is intentional and matches the single-line behavior asserted by
+    /// <see cref="VerifyOnlyFirstDocumentationCommentIsReplaced"/>: the analyzer flags only the first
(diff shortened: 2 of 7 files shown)


Reihitsu.Analyzer.Test/Documentation/RH8201InheritdocShouldBeUsedAnalyzerTests.cs
```diff
@@ -408,7 +407,294 @@ public override void TestMethod()
...
+                                        public override void TestMethod()
+                                        {
+                                        }
+                                    }
+                                }
+                                """;
+
+        var fixedSource = await ApplyCodeFixAsync(NormalizeToCarriageReturnLineFeed(testData));
+
+        Assert.Contains("/// <inheritdoc/>\r\n        public override void TestMethod()", fixedSource);
```
thoenissen, line 703: The fix now deletes a line break in the trivia gap between the replaced `/** */` comment and the member, but no test puts a comment or a preprocessor directive in that gap — the two shapes the repo's test expectations require for any rule that removes a line break. Both currently behave correctly (verified locally), so this is a coverage gap that leaves the deletion unguarded against regressions. Add:

```csharp
    /** <summary>doc</summary> */ // trailing
    public override void TestMethod()
```

```csharp
    /** <summary>doc</summary> */
#if DEBUG
    public override void TestMethod()
    {
    }
#endif
```

RH8201 also has no FixAll test, although several `/** */`-documented overrides in one type is the common case; add one alongside these.

---

Reihitsu.Analyzer.CodeFixes/Rules/Documentation/RH8201InheritdocShouldBeUsedCodeFixProvider.cs
```diff
@@ -43,36 +43,79 @@ private static SyntaxTrivia CreateInheritdocTrivia(string endOfLine)
...
+
+                pendingTrivia.Clear();
+            }
+
             // Only the flagged (first) documentation comment is replaced. Replacing every documentation comment would
             // emit multiple <inheritdoc/> lines when a member carries more than one documentation comment
             if (replaced == false
-                && trivia.IsKind(SyntaxKind.SingleLineDocumentationCommentTrivia))
+                && (trivia.IsKind(SyntaxKind.SingleLineDocumentationCommentTrivia)
+                    || trivia.IsKind(SyntaxKind.MultiLineDocumentationCommentTrivia)))
```
thoenissen: This is a second private copy of the analyzer's documentation-comment-kind predicate (`RH8201InheritdocShouldBeUsedAnalyzer.cs:50-51`), and the drift between those two copies is exactly the bug this PR fixes — nothing prevents it from recurring. The repository already carries five more inline copies (`RH5101...CodeFixProvider.cs:52`, `RH5102...CodeFixProvider.cs:52`, `RH5111...CodeFixProvider.cs:96`, `RH5204...CodeFixProvider.cs:193`, `BlankLineTokenCleanupRewriter.cs:296`). Extract `SyntaxTriviaUtilities.IsDocumentationCommentTrivia` in `Reihitsu.Core` and call it from both the analyzer and this fix so analyzer/fix parity is structural instead of coincidental.

---

Reihitsu.Analyzer.CodeFixes/Rules/Documentation/RH8201InheritdocShouldBeUsedCodeFixProvider.cs
```diff
@@ -43,36 +43,79 @@ private static SyntaxTrivia CreateInheritdocTrivia(string endOfLine)
...
 
         foreach (var trivia in triviaList)
         {
+            if (isLineBreakPending)
+            {
+                // Whitespace between the replaced comment and its line break is buffered so it can be restored
+                // when no line break follows after all
+                if (trivia.IsKind(SyntaxKind.WhitespaceTrivia))
+                {
+                    pendingTrivia.Add(trivia);
```
thoenissen: The whitespace buffering has no observable effect. I disabled it locally (dropping the buffered whitespace instead of restoring it) and both shapes that can reach it produce byte-identical fixed documents, because `CodeAction.GetOperationsAsync` post-processing re-indents the affected lines:

- `/** <summary>d</summary> */ public override void TestMethod()` → `    /// <inheritdoc/>` + correctly indented member
- `/** <summary>d</summary> */ // trailing` → `    /// <inheritdoc/>` + correctly indented `// trailing`

A directive cannot follow `*/` on the same line, so no third shape reaches this branch. Drop `pendingTrivia`, the in-loop flush (lines 80-85) and the post-loop flush (lines 109-112) and keep only the line-break suppression — that also removes the buffering state from `ReplaceDocumentation`, which currently mixes replacement with a trivia state machine. The line-break drop itself is required: with `isLineBreakPending` forced to `false` the own-line form gains a blank line before the member, so keep that part.

---

thoenissen (commented):
The core change is correct: `MultiLineDocumentationCommentTrivia` is now matched, the analyzer's `FirstOrDefault` selection and the fix's selection agree again, and dropping the terminating line break is genuinely required — forcing `isLineBreakPending` to `false` reintroduces a blank line before the member on the own-line `/** */` form. Directive, trailing-comment and same-line shapes all produce correct output.

Three findings are attached inline: missing gap tests (comment, directive, FixAll), a duplicated documentation-comment-kind predicate that keeps the analyzer/fix drift class open, and a whitespace-buffering branch that has no observable effect on the produced document.

---

Reihitsu.Analyzer.Test/Documentation/RH8201InheritdocShouldBeUsedAnalyzerTests.cs
thoenissen: Addressed: added the two gap shapes and a Fix All test (959ff70).

- `VerifyTrailingCommentIsPreservedWhenMultiLineDocumentationCommentIsReplaced` — `/** ... */ // trailing`; the trailing comment keeps its own line.
- `VerifyDirectiveIsPreservedWhenMultiLineDocumentationCommentIsReplaced` — directive between the comment and the member. Used `#if true` rather than `#if DEBUG`: the test verifier parses without preprocessor symbols, so under `#if DEBUG` the override is disabled text, the doc comment attaches to nothing and RH8201 never fires.
- `VerifyMultipleMultiLineDocumentationCommentsAreFixedInOneFixAllIteration` — two `/** */`-documented overrides in one type, `NumberOfFixAllIterations = 1`.

---

Reihitsu.Analyzer.CodeFixes/Rules/Documentation/RH8201InheritdocShouldBeUsedCodeFixProvider.cs
thoenissen: Addressed: extracted `SyntaxTriviaUtilities.IsDocumentationCommentTrivia` in `Reihitsu.Core` and called it from both the analyzer (`RH8201InheritdocShouldBeUsedAnalyzer.cs:49`) and this fix (959ff70). `IsCommentTrivia` now delegates to it instead of repeating the two kinds, and four focused tests were added in `SyntaxTriviaUtilitiesTests`.

Packaging is safe — `Reihitsu.Core.dll` is already packed under `analyzers/dotnet/cs` and both projects already referenced Core.

The five other inline copies you listed (`RH5101`/`RH5102`/`RH5111`/`RH5204` code fixes, `BlankLineTokenCleanupRewriter`) are untouched: they test `IsCommentTrivia`-shaped sets rather than the documentation-only predicate, so folding them in is a wider change than this fix. Still open on this PR if you want it done here.

---

Reihitsu.Analyzer.CodeFixes/Rules/Documentation/RH8201InheritdocShouldBeUsedCodeFixProvider.cs
thoenissen: Addressed: dropped `pendingTrivia` and both flushes, kept the line-break suppression (959ff70). `ReplaceDocumentation` no longer carries buffering state — whitespace before the line break is now skipped rather than buffered.

Confirmed independently rather than taken on trust: I added the trailing-comment and directive gap tests first and ran them against the buffered implementation, then applied this simplification and re-ran. Identical results both times (16/16), including `/** ... */ // trailing`, whose expectation is the re-indented `// trailing` — so the buffering was indeed unobservable.

---

Reihitsu.Analyzer.Test/Documentation/RH8201InheritdocShouldBeUsedAnalyzerTests.cs
thoenissen: Confirmed fixed. All three tests are present and the gap shapes are covered; `#if true` is the right call, since the verifier parses without preprocessor symbols. Ran `Reihitsu.Analyzer.Test` filtered to `RH8201` on `7d99737`: 16/16 pass.

---

Reihitsu.Analyzer.CodeFixes/Rules/Documentation/RH8201InheritdocShouldBeUsedCodeFixProvider.cs
thoenissen: Confirmed fixed. `pendingTrivia` and both flushes are gone, the line-break suppression is kept, and `ReplaceDocumentation` no longer carries buffering state. Ran `Reihitsu.Analyzer.Test` filtered to `RH8201` on `7d99737`: 16/16 pass, including the trailing-comment expectation.

---

Reihitsu.Analyzer.CodeFixes/Rules/Documentation/RH8201InheritdocShouldBeUsedCodeFixProvider.cs
thoenissen: The parity part is fixed: `SyntaxTriviaUtilities.IsDocumentationCommentTrivia` is now the single source for the analyzer and the fix, and `Reihitsu.Core.dll` is indeed packed under `analyzers/dotnet/cs` (`Reihitsu.Analyzer.Package.csproj:47`).

My list of five was wrong on four of them — `RH5101`, `RH5102`, `RH5111` and `RH5204` are `IsCommentTrivia`-shaped, as you say, so leave them.

Three genuine copies of the extracted predicate survive, all in `Reihitsu.Formatter`, which already references `Reihitsu.Core`:

- `BlankLineTokenCleanupRewriter.cs:293` — a private `IsDocumentationCommentTrivia` with a body identical to the new Core helper, one call site
- `BlankLineTokenCleanupRewriter.cs:44` — the same two kinds inline in `HasDocumentationCommentInLeadingTrivia`
- `BlankLineCollapser.cs:101` — the same two kinds inline, negated

Replace all three with the Core helper to close the class.

---

thoenissen (commented):
Re-review of `7d99737`. Two of the three prior findings are confirmed fixed and their threads are resolved; the duplication thread stays open with a narrowed scope (three genuine copies of the extracted predicate survive in `Reihitsu.Formatter`) — see the reply there.

One new finding, no line anchor:

1. The PR description no longer matches the code. It still states "whitespace following the comment is buffered and restored when no line break follows it, which keeps the odd `/** doc */ public override ...` single-line layout compiling" — `959ff70` removed that buffering; the whitespace is now dropped and the code action's post-processing re-indents. It also states "Six analyzer tests cover the new behavior"; there are now nine, plus four `SyntaxTriviaUtilities` tests in `Reihitsu.Core.Test`. Update both bullets so the description describes the merged change.

---

Reihitsu.Analyzer.CodeFixes/Rules/Documentation/RH8201InheritdocShouldBeUsedCodeFixProvider.cs
thoenissen: Addressed: all three formatter copies now call `SyntaxTriviaUtilities.IsDocumentationCommentTrivia` (50fa82a).

- `BlankLineTokenCleanupRewriter.cs:44` — `HasDocumentationCommentInLeadingTrivia` passes the Core helper as the predicate directly.
- `BlankLineTokenCleanupRewriter.cs:293` — the private copy is deleted and its single call site now calls the Core helper.
- `BlankLineCollapser.cs:101` — `EndsWithEmbeddedLineBreak` now uses `SyntaxTriviaUtilities` for both halves of the check.

No behavior change: all three bodies tested the same two syntax kinds. Formatter suite green (1018/1018), full validation green across all four projects.

---

thoenissen: Confirmed fixed. All three formatter call sites now use the Core helper, and the substitutions are exactly equivalent — `EndsWithEmbeddedLineBreak` is the De Morgan form of the original `!Single && !Multi`. Grepped the repository: `SyntaxTriviaUtilities.IsDocumentationCommentTrivia` is now the only two-kind documentation predicate, so the class is closed. Built the solution on `c6a7173` (clean, one pre-existing CS1574) and ran `Reihitsu.Formatter.Test` filtered to `BlankLine`: 113/113 pass.

---

thoenissen (commented):
Re-review of `c6a7173`. All three inline threads are now confirmed fixed and resolved. Two items remain, neither with a changed-line anchor:

1. **Still open from the previous round** — the PR description has not been updated. It still states "whitespace following the comment is buffered and restored when no line break follows it", which `959ff70` removed, and "Six analyzer tests", where there are now nine plus four `SyntaxTriviaUtilities` tests.

2. **New, low** — the same duplication one level up. `SyntaxTriviaUtilities.IsCommentTrivia` is the canonical four-kind comment predicate, yet eight inline copies of it survive, two of them inside `Reihitsu.Core` itself: `SyntaxNodeUtilities.cs:42` (`IsComment`), `UsingDirectiveOrderingUtilities.cs:317` (`HasCommentTrivia`), `CommentIndentationContributor.cs:58` (`IsComment`), `RH5204IndentationMustUseFourSpacesPerScopeLevelAnalyzer.cs:277` (`IsComment`), `RH5204IndentationMustUseFourSpacesPerScopeLevelCodeFixProvider.cs:188` (`IsCommentTrivia`), `RH5101FirstArgumentShouldBeOnSameLineCodeFixProvider.cs:48`, `RH5102ArgumentsShouldBeOnSingleOrSeparateLinesCodeFixProvider.cs:48`, `RH5111AssignmentsMustHaveProperLineBreaksCodeFixProvider.cs:92` and `RH7310EmptyRegionsShouldBeRemovedAnalyzer.cs:83`. This corrects my earlier reply, which dismissed four of these as out of scope — they are not copies of the documentation predicate, but they are copies of `IsCommentTrivia`. Route them through the Core helper as well.

The documentation-predicate class itself is fully closed: a repository-wide grep finds no remaining two-kind copy.

---

PR #566, opened 2026-07-25
Restrict the conditional-access chain walk to the chain spine

thoenissen:
## Summary

`ChainWalker.IsInsideConditionalAccess` now follows the access-chain spine only instead of walking every parent up to the enclosing statement, so an invocation that merely sits inside an argument or a lambda body of a conditional-access chain is no longer treated as a link of that chain. `MethodChainAlignmentContributor` drops its duplicated copy of the check and reuses the shared helper.

Following review, a comment-exempt chain — one a preceding comment keeps wrapped — now takes its continuation column from the chain root token, so its links line up directly under the root instead of relative to the enclosing statement.

## Why

A wrapped chain continuation nested in a lambda argument lost its indentation and was dedented to method-body level. The alignment contributor skipped the inner invocation because the old walk crossed the argument and lambda boundaries and found the outer `?.` chain, so no contributor claimed the continuation line and only pass-1 block indentation was left on it.

The comment-exempt path had the same failure mode from a different direction: with no first dot to align against, it fell back to the continuation line's block indentation, which loses the chain whenever the root sits far into the line.

## Linked issues

Closes #475

## Review notes

The corrected walk also feeds `ChainWalker.IsOutermostChainInvocation`, which the line-break phase uses, so `CollapseChainToSingleLine` and `NormalizeChain` now run for chains nested in an argument or lambda of a conditional-access chain. Both paths have regression tests in `Regression/LineBreaks/ChainInsideConditionalAccessArgumentTests.cs`.

`IsChainSpineNode` is private to its single caller and `IsOutermostChainNode` keeps its own explicit node list, so widening one walk cannot silently widen the other.

The comment-exempt column change is a behavior change beyond the reported shape: it also moves plain statement-level chains, and the two fixtures that pinned the old statement-relative column were updated to the root-aligned one. The comment-exempt anchor is the root token while the normal path anchors on the first dot; the two columns differ for the same chain, which is deliberate — a comment-exempt chain has no first dot on the root line to align against.

**Known gap, not addressed here.** When a directive sits in the chain gap, a single-link chain still keeps pass-1 block indentation, because it collects one alignment dot and `Contribute` returns at `dots.Count < 2`. This is not specific to conditional compilation — `#if`/`#else` and `#pragma warning disable` in that position both produce the same column-8 continuation. It reproduces byte-identically on the base commit and without the `?.`, so it is independent of this change; closing it would mean defining a continuation column for single-link wrapped chains, which nothing currently pins.

## Follow-up work

- Define the continuation column for single-link chains that cannot be rejoined, covering any directive the rejoin refuses to join across — not `#if` alone.

### DIFF
--- a/Reihitsu.Formatter.Test/Regression/Indentation/MethodChainAlignmentTests.cs
+++ b/Reihitsu.Formatter.Test/Regression/Indentation/MethodChainAlignmentTests.cs
@@ -302,7 +302,8 @@ public class MethodChainAlignmentTests : FormatterTestsBase
     }
 
     /// <summary>
-    /// Verifies that a chain with a comment directly above the first wrapped call remains unchanged
+    /// Verifies that a chain with a comment directly above the first wrapped call keeps its links
+    /// aligned under the chain root token
     /// </summary>
     [TestMethod]
     public void ChainWithCommentAboveFirstWrappedCallRemainsUnchanged()
@@ -318,9 +319,9 @@ public class MethodChainAlignmentTests : FormatterTestsBase
         const string expected = """
                                 var x = a
 
-                                    // Keep this step separate.
-                                    .Foo()
-                                    .Bar();
+                                        // Keep this step separate.
+                                        .Foo()
+                                        .Bar();
                                 """;
 
         // Act & Assert
@@ -972,6 +973,32 @@ public class MethodChainAlignmentTests : FormatterTestsBase
                                                                      return true;
                                                                  }
                                                  });
+                                 }
+                             }
+                             """;
+
+        // Act & Assert
+        AssertRuleResult(input);
+    }
+
+    /// <summary>
+    /// Verifies that a comment-exempt chain nested in a lambda argument of a conditional access
+    /// chain aligns its links under the chain root instead of the enclosing block (issue #475)
+    /// </summary>
+    [TestMethod]
+    public void CommentExemptChainInsideLambdaArgumentAlignsUnderChainRoot()
+    {
+        // Arrange
+        const string input = """
+                             public class Class
+                             {
+                                 public void M()
+                                 {
+                                     var value = a.List.Find(e => d
+
+                                                                  // Keep this call separate.
+                                                                  .Equals(e.Number))
+                                                       ?.Value;
                                  }
                              }
                              """;
--- a/Reihitsu.Formatter.Test/Regression/Indentation/MethodChainAlignmentTests.cs
+++ b/Reihitsu.Formatter.Test/Regression/Indentation/MethodChainAlignmentTests.cs
@@ -980,5 +980,94 @@ public class MethodChainAlignmentTests : FormatterTestsBase
         AssertRuleResult(input);
     }
 
+    /// <summary>
+    /// Verifies that a misaligned chain continuation inside a lambda argument is aligned to the
+    /// chain anchor when the outer chain is a conditional access expression (issue #475)
+    /// </summary>
+    [TestMethod]
+    public void MisalignedChainInsideLambdaArgumentOfConditionalAccessIsAligned()
+    {
+        // Arrange
+        const string input = """
+                             public class Class
+                             {
+                                 public void M()
+                                 {
+                                     var value = a.List.Find(e => d.Equals(e.Number.ToString()
+                                         .Trim(),
+                                                                           StringComparison.OrdinalIgnoreCase))
+                                                       ?.Value;
+                                 }
+                             }
+                             """;
+
+        const string expected = """
+                                public class Class
+                                {
+                                    public void M()
+                                    {
+                                        var value = a.List.Find(e => d.Equals(e.Number.ToString()
+                                                                                      .Trim(),
+                                                                              StringComparison.OrdinalIgnoreCase))
+                                                          ?.Value;
+                                    }
+                                }
+                                """;
+
+        // Act & Assert
+        AssertRuleResult(input, expected);
+    }
+
+    /// <summary>
+    /// Verifies that a wrapped chain continuation inside a plain argument keeps its alignment
+    /// when the outer chain is a conditional access expression (issue #475)
+    /// </summary>
+    [TestMethod]
+    public void ChainInsideArgumentOfConditionalAccessKeepsAlignment()
+    {
+        // Arrange
+        const string input = """
+                             public class Class
+                             {
+                                 public void M()
+                                 {
+                                     var value = a.List.Find(d.Equals(e.Number.ToString()
+                                                                              .Trim()))
+                                                       ?.Value;
+                                 }
+                             }
+                             """;
+
+        // Act & Assert
+        AssertRuleResult(input);
+    }
+
+    /// <summary>
+    /// Verifies that a wrapped chain continuation inside a lambda argument keeps its alignment
+    /// when the outer chain is a conditional access expression (issue #475)
+    /// </summary>
+    [TestMethod]
+    public void ChainInsideLambdaArgumentOfConditionalAccessKeepsAlignment()
+    {
+        // Arrange
+        const string input = """
+                             public class Class
+                             {
+                                 public void M()
+                                 {
+                                     var a = new A();
+                                     var d = string.Empty;
+                                     var value = a.List.Find(e => d.Equals(e.Number.ToString()
+                                                                                   .Trim(),
+                                                                           StringComparison.OrdinalIgnoreCase))
+                                                       ?.Value;
+                                 }
+                             }
+                             """;
+
+        // Act & Assert
+        AssertRuleResult(input);
+    }
+
     #endregion // Methods
 }
--- a/Reihitsu.Formatter.Test/Regression/Indentation/MethodChainAlignmentTests.cs
+++ b/Reihitsu.Formatter.Test/Regression/Indentation/MethodChainAlignmentTests.cs
@@ -306,7 +306,7 @@ public class MethodChainAlignmentTests : FormatterTestsBase
     /// aligned under the chain root token
     /// </summary>
     [TestMethod]
-    public void ChainWithCommentAboveFirstWrappedCallRemainsUnchanged()
+    public void ChainWithCommentAboveFirstWrappedCallAlignsUnderChainRoot()
     {
         // Arrange
         const string input = """
--- a/Reihitsu.Formatter/Pipeline/Indentation/Contributors/MethodChainAlignmentContributor.cs
+++ b/Reihitsu.Formatter/Pipeline/Indentation/Contributors/MethodChainAlignmentContributor.cs
@@ -102,6 +102,21 @@ internal sealed class MethodChainAlignmentContributor : ILayoutContributor
         }
     }
 
+    /// <summary>
+    /// Computes the continuation column for a chain that a preceding comment keeps wrapped. Such a
+    /// chain has no first dot to align against, so the continuation dots line up with the chain root
+    /// token itself. Measuring from the root rather than from the continuation line's block
+    /// indentation keeps the chain under its root even when the root sits far into the line, for
+    /// example inside an argument or a lambda body
+    /// </summary>
+    /// <param name="node">The chain node being laid out</param>
+    /// <param name="model">The layout model</param>
+    /// <returns>The column for the chain's continuation lines</returns>
+    private static int GetCommentExemptContinuationColumn(SyntaxNode node, LayoutModel model)
+    {
+        return LayoutComputer.GetAdjustedColumn(node.GetFirstToken(), model);
+    }
+
     #endregion // Methods
 
     #region ILayoutContributor
@@ -118,7 +133,7 @@ internal sealed class MethodChainAlignmentContributor : ILayoutContributor
 
         if (ShouldKeepFirstWrappedCallOnContinuationLine(dots[0]))
         {
-            var continuationColumn = LayoutComputer.GetAdjustedColumn(dots[0], model) + FormattingContext.IndentSize;
+            var continuationColumn = GetCommentExemptContinuationColumn(node, model);
 
             foreach (var dot in dots)
             {
--- a/Reihitsu.Formatter/Pipeline/Indentation/Contributors/MethodChainAlignmentContributor.cs
+++ b/Reihitsu.Formatter/Pipeline/Indentation/Contributors/MethodChainAlignmentContributor.cs
@@ -102,34 +102,6 @@ internal sealed class MethodChainAlignmentContributor : ILayoutContributor
         }
     }
 
-    /// <summary>
-    /// Determines whether a syntax node is nested inside a <see cref="ConditionalAccessExpressionSyntax"/>.
-    /// Walks up the parent chain until a statement or member declaration is found
-    /// </summary>
-    /// <param name="node">The node to check</param>
-    /// <returns><see langword="true"/> if the node is inside a conditional access expression; otherwise, <see langword="false"/></returns>
-    private static bool IsInsideConditionalAccess(SyntaxNode node)
-    {
-        var current = node.Parent;
-
-        while (current != null)
-        {
-            if (current is ConditionalAccessExpressionSyntax)
-            {
-                return true;
-            }
-
-            if (current is StatementSyntax || current is MemberDeclarationSyntax)
-            {
-                return false;
-            }
-
-            current = current.Parent;
-        }
-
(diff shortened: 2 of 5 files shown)


Reihitsu.Formatter/Pipeline/Indentation/Contributors/MethodChainAlignmentContributor.cs
```diff
@@ -240,7 +212,7 @@ private static bool ShouldSkipInvocation(InvocationExpressionSyntax invocation)
             return true;
         }
 
-        return IsInsideConditionalAccess(invocation);
+        return ChainWalker.IsInsideConditionalAccess(invocation);
```
thoenissen, line 230: Un-skipping conditional-access-nested invocations routes single-link chains into `ShouldKeepFirstWrappedCallOnContinuationLine`, which derives the continuation column from the dot's own pass-1 block indentation (`GetAdjustedColumn(dots[0], model) + FormattingContext.IndentSize`). For a chain root nested in a lambda argument that resolves to method-body + 4, so #475's dedent symptom survives for this shape. Input:

```csharp
        var value = a.List.Find(e => d
                                      // keep wrapped
                                      .Equals(e.Number))
                          ?.Value;
```

Head output (`d` is at column 37, the continuation lands at column 12):

```csharp
        var value = a.List.Find(e => d

            // keep wrapped
            .Equals(e.Number))
                          ?.Value;
```

Derive the continuation column from the chain root token instead of from the dot's block indentation.

---

Reihitsu.Formatter/Pipeline/LineBreaks/ChainWalker.cs
```diff
@@ -185,7 +200,7 @@ public static bool IsInsideConditionalAccess(SyntaxNode node)
                 return true;
             }
 
-            if (current is StatementSyntax || current is MemberDeclarationSyntax)
+            if (IsChainSpineNode(current) == false)
```
thoenissen, line 191: The corrected walk does not close the defect class when a preprocessor directive sits in the gap. The rejoin is correctly refused there, but a single-link inner chain collects one alignment dot only, so `Contribute` returns at `dots.Count < 2` and the continuation keeps pass-1 block indentation — the same #475 symptom. Input:

```csharp
        var value = a.List.Find(e => d
#if DEBUG
                                      .Equals(e.Number))
#else
                                      .Equals(e.Other))
#endif
                          ?.Value;
```

Head output, byte-identical to the base commit (the active `#else` link is dedented to column 8):

```csharp
        var value = a.List.Find(e => d
#if DEBUG
                                      .Equals(e.Number))
#else
        .Equals(e.Other))
#endif
                          ?.Value;
```

The root cause is separate from the walk (it reproduces without the `?.` as well), but the reported symptom stays open. Contribute an anchor for single-dot chains whose root is not first on its line, or state the gap in the PR description.

---

Reihitsu.Formatter.Test/Regression/Indentation/MethodChainAlignmentTests.cs
```diff
@@ -980,5 +980,94 @@ void Build()
...
+        // Act & Assert
+        AssertRuleResult(input);
+    }
+
+    /// <summary>
+    /// Verifies that a wrapped chain continuation inside a lambda argument keeps its alignment
+    /// when the outer chain is a conditional access expression (issue #475)
+    /// </summary>
+    [TestMethod]
+    public void ChainInsideLambdaArgumentOfConditionalAccessKeepsAlignment()
```
thoenissen, line 1077: All three new tests exercise the alignment anchor only. `IsInsideConditionalAccess` also feeds `IsOutermostChainInvocation`, so `ChainLineBreakRewriter.VisitInvocationExpression` now runs `IsCollapsibleChain`/`CollapseChainToSingleLine`/`NormalizeChain` on chains nested in an argument or lambda of a conditional-access chain — a path that never ran for those shapes. That output change is untested. Input:

```csharp
        var value = a.List.Find(e => d
                                      .Equals(e.Number))
                          ?.Value;
```

Base keeps the chain wrapped; head rejoins it:

```csharp
        var value = a.List.Find(e => d.Equals(e.Number))
                          ?.Value;
```

Add a line-break regression test for the collapse and the normalize path in this nesting.

---

Reihitsu.Formatter/Pipeline/LineBreaks/ChainWalker.cs
```diff
@@ -161,16 +161,31 @@ public static bool IsOutermostChainInvocation(InvocationExpressionSyntax node)
...
 
     /// <summary>
-    /// Determines whether a syntax node is nested inside a <see cref="ConditionalAccessExpressionSyntax"/>
+    /// Determines whether a syntax node is a link on an access chain spine. Only the spine positions
+    /// count, so a node reached through an argument list, lambda body, or other nested construct is
+    /// not a spine link
+    /// </summary>
+    /// <param name="node">The node to check</param>
+    /// <returns><see langword="true"/> if the node is a chain spine link; otherwise, <see langword="false"/></returns>
+    public static bool IsChainSpineNode(SyntaxNode node)
```
thoenissen: The summary promises positional semantics the method does not implement: it is a pure type test and returns `true` for any `InvocationExpressionSyntax`, including one that sits inside an argument list or a lambda body. Spine membership is established by the caller's parent walk, not here. Reword the doc to describe the type test and name the caller's walk as the source of the positional guarantee, so a future caller does not rely on a check that is not performed.

---

thoenissen (commented):
The reported #475 shape is fixed and the change is idempotent — a second CLI pass over ten fixtures reported no changes. The narrowed walk is a strict subset of the old one, so nothing that was previously aligned loses its anchor, and it now matches `FluentChainAnalysisHelper.IsInnerChainMember`, which already walked the spine only. Line-anchored findings are in the inline comments. Two findings have no changed-line anchor:

1. **Duplicated chain-spine policy across assemblies.** `Reihitsu.Analyzer/Core/FluentChainAnalysisHelper.IsInnerChainMember` re-implements the walk this PR just consolidated in the formatter: continue up through invocation/element-access/postfix, stop at anything else. The two node sets already differ — `MemberBindingExpressionSyntax` is in `ChainWalker.IsChainSpineNode` but not in the analyzer's switch. RH5201/RH5112 and the indentation and line-break phases must agree on this policy, so it belongs in `Reihitsu.Core` next to `FluentChainUtilities` with one definition, not one per assembly.

2. **Pre-existing: the blank-lines phase splits a comment-preceded chain continuation.** Independent of this diff (base and head produce identical output), but it corrupts every fixture in the area this PR touches. Input `var value = a` / `// keep wrapped` / `.Foo()` / `.Bar();` gains a blank line between `a` and the comment, so the expression is broken in two. It reproduces at statement level and inside an argument list, and it is stable across passes, so no test catches it. Affected: `Reihitsu.Formatter/Pipeline/BlankLines`.

---

Reihitsu.Formatter/Pipeline/Indentation/Contributors/MethodChainAlignmentContributor.cs
thoenissen: Addressed, with a different column than proposed (337db9e).

A comment-exempt chain has no first dot to align against, so the fallback to the continuation line's block indentation placed the links relative to the enclosing statement. That is what loses the chain when the root sits inside an argument or a lambda body.

The continuation column is now measured from the chain root token, and the links sit directly under the root rather than one indent step in from it — for this shape the dot lands under `d`. Deriving it as root + one indent step was tried first and rejected: it put the links one step past the root in every case, including plain statement-level chains.

This shape had no defined behavior before, so the two fixtures that pinned the old statement-relative column were updated to the root-aligned one, and a case nested in a lambda argument was added.

---

Reihitsu.Formatter.Test/Regression/Indentation/MethodChainAlignmentTests.cs
thoenissen: Addressed (5955b9e). Added `Regression/LineBreaks/ChainInsideConditionalAccessArgumentTests.cs` covering both newly reachable paths in this nesting: `CollapseChainToSingleLine` rejoining a collapsible chain onto the root line, and `NormalizeChain` moving a trailing link onto its own line while the chain stays wrapped because it has an intermediate member access. Both outputs were confirmed to differ from the base commit.

---

Reihitsu.Formatter/Pipeline/LineBreaks/ChainWalker.cs
thoenissen: Confirmed and recorded as a known gap rather than fixed here, per the second option offered.

Verified against the base commit: the output is byte-identical, so the walk change neither causes nor worsens it. As noted, the root cause is separate — the single-link chain collects one alignment dot, `Contribute` returns at `dots.Count < 2`, and it reproduces without the `?.`.

Closing it means defining a continuation column for single-link wrapped chains, which nothing currently pins, so it is left out of this PR to keep the change scoped to the spine walk. The gap is stated in the PR description.

---

Reihitsu.Formatter/Pipeline/LineBreaks/ChainWalker.cs
thoenissen: Addressed (5955b9e). The summary now states it is a type test only, notes that a node of the same kind is just as reachable through an argument list or a lambda body, and names `IsInsideConditionalAccess`'s parent walk as the source of the positional guarantee.

The method is also `private` now with that single caller, so a future caller cannot pick it up expecting spine membership. `IsOutermostChainNode` was reverted to its own explicit node list for the same reason — sharing the helper coupled two questions that only happen to agree today.

---

Reihitsu.Formatter/Pipeline/Indentation/Contributors/MethodChainAlignmentContributor.cs
thoenissen: Verified fixed. `GetCommentExemptContinuationColumn` measures from `node.GetFirstToken()`, and the lambda-argument shape now lands at column 37, directly under `d`. Aligning under the root rather than root + one indent step is coherent — RH5201 only requires the links to agree with the first one, so the absolute column is the formatter's to define.

Also checked the pre-order hazard that `GetChainAnchorColumn` works around for initializers: a root on an object-initializer close brace, a root inside an initializer value, and a root after a multi-line argument list all resolve to the root column, and a second CLI pass over all of them is a no-op.

---

Reihitsu.Formatter.Test/Regression/Indentation/MethodChainAlignmentTests.cs
thoenissen: Verified. Both new tests match the observed head behavior: the collapsible chain rejoins to `a.List.Find(e => d.Equals(e.Number))`, and the intermediate-member-access chain stays wrapped with the trailing link moved onto its own line. All 59 chain tests pass at f7124fd.

---

Reihitsu.Formatter/Pipeline/LineBreaks/ChainWalker.cs
thoenissen: Verified. The summary now scopes itself to a type test and names the caller's walk as the source of the positional guarantee, the method is `private` to its single caller, and `IsOutermostChainNode` is back on its own explicit node list, so widening one walk can no longer widen the other.

---

Reihitsu.Formatter/Pipeline/LineBreaks/ChainWalker.cs
thoenissen: The documented-gap route is accepted; the base-commit comparison matches. One addition for the follow-up: the gap is not specific to conditional compilation. A `#pragma warning disable` in the same position produces the same column-8 continuation, so the follow-up has to cover any directive the rejoin refuses to join across, not `#if` alone.

Leaving this thread open until that lands.

---

Reihitsu.Formatter.Test/Regression/Indentation/MethodChainAlignmentTests.cs
```diff
@@ -302,7 +302,8 @@ public void ChainWithFirstLinkOnDifferentLineCollapsesAndAligns()
     }
 
     /// <summary>
-    /// Verifies that a chain with a comment directly above the first wrapped call remains unchanged
+    /// Verifies that a chain with a comment directly above the first wrapped call keeps its links
+    /// aligned under the chain root token
```
thoenissen, line 306: The summary was corrected here but the method name on line 309 still says `RemainsUnchanged`, while the test asserts a transformation — input at column 0, expected at column 8. Rename it to match the new contract, for example `ChainWithCommentAboveFirstWrappedCallAlignsUnderChainRoot`.

---

thoenissen (commented):
Re-review of f7124fd. Three of the four inline findings are verified fixed and resolved; the directive-gap thread stays open as the documented follow-up. The comment-exempt column change holds up under the shapes it newly reaches — roots on an initializer close brace, inside an initializer value, after a multi-line argument list, nested comment-exempt chains, and CRLF input all resolve to the root column, and a second CLI pass over every fixture is a no-op. One new finding is inline. Two items have no changed-line anchor:

1. **Retraction of a finding from the previous review.** The blank line inserted before a comment-preceded chain continuation is not a defect — RH5020 requires single-line comments to be preceded by a blank line, and the committed `MethodChainAlignmentFullPipelineTests` fixture already pins it. The earlier review reported it as a pre-existing blank-lines bug; that was wrong, and no action is needed. The region-directive removal seen in the same area is likewise policy, implemented by `NestedRegionRemovalStep`.

2. **Still open: duplicated chain-spine policy across assemblies.** `Reihitsu.Analyzer/Core/FluentChainAnalysisHelper.IsInnerChainMember` still re-implements the walk this PR consolidated in the formatter — continue up through invocation/element-access/postfix, stop at anything else — and the node sets still differ, with `MemberBindingExpressionSyntax` present only in `ChainWalker.IsChainSpineNode`. Making that helper `private` narrows the formatter's exposure but leaves the cross-assembly copy in place. RH5201, RH5112, the indentation phase, and the line-break phase all depend on this policy agreeing, so it belongs in `Reihitsu.Core` with one definition.

3. **Low: `documentation/rules/RH5112.md` no longer matches the formatter.** Line 22 reads "If there is a comment directly above the first wrapped call, keep that call layout as written," but the formatter now re-columns that call to the chain root — the updated unit-test fixture moves it from column 0 to column 8. Reword it to state that the wrap is preserved while the continuation column is normalized to the chain root, so the documented contract matches the behavior this PR defines.

---

Reihitsu.Formatter.Test/Regression/Indentation/MethodChainAlignmentTests.cs
thoenissen: Addressed (80795e0). Renamed to `ChainWithCommentAboveFirstWrappedCallAlignsUnderChainRoot`.

`CommentAboveFirstWrappedCallKeepsChainUnchanged` in `Regression/FullPipeline/MethodChainAlignmentFullPipelineTests.cs` had the same stale contract — its input sits at column 12 and its expected at 15 — so it was renamed to `CommentAboveFirstWrappedCallAlignsChainUnderRoot` and its summary corrected as well.

---

Reihitsu.Formatter/Pipeline/LineBreaks/ChainWalker.cs
thoenissen: Confirmed and folded into the note. Reproduced `#pragma warning disable CS0618` in the same position and it produces the identical column-8 continuation, so the gap is not conditional-compilation specific.

The PR description now states that both `#if`/`#else` and `#pragma warning disable` trigger it, and the follow-up item asks for any directive the rejoin refuses to join across rather than `#if` alone.

Resolving this thread: the remaining work is the follow-up itself, which is out of scope for this PR by the documented-gap decision.

---

Reihitsu.Formatter.Test/Regression/Indentation/MethodChainAlignmentTests.cs
thoenissen: Verified. Both names now match their assertions, each resolves to a single declaration, and no other chain test carries a stale unchanged-contract name — `SingleLineChainRemainsUnchanged` and `MethodChainInForeachDeclarationRemainsUnchanged` both genuinely assert an unchanged input.

---

PR #567, opened 2026-07-26
Brace multi-line switch sections ending in return or throw

thoenissen:
## Summary

Recognize switch sections as multi-line when a non-terminal statement is followed by `return` or `throw` on a later line, then brace non-fall-through sections consistently. Preserve cross-section `goto` targets, add LF/CRLF second-pass coverage for both terminal forms, and update the affected self-hosted analyzer source.

## Why

Terminal statements were omitted from the section-size decision, causing assignment-plus-return bodies to remain unbraced despite spanning multiple lines. Brace insertion also needs to preserve valid control flow when an ordinary `goto` targets a label owned by another switch section.

## Linked issues

Closes #476

## Review notes

A lone multi-line return and `statement + break` sections retain their existing brace-free behavior. Switches with cross-section label jumps are left unchanged; jumps inside nested switches are included when they target an outer-section label, while `goto case` and `goto default` are unaffected. Cross-section lookup ignores labels and jumps inside local functions and lambdas because they have independent executable scopes.

## Follow-up work

None.

### DIFF
--- a/Reihitsu.Formatter.Test/Unit/SwitchCaseBraces/SwitchCaseBraceRewriterTests.cs
+++ b/Reihitsu.Formatter.Test/Unit/SwitchCaseBraces/SwitchCaseBraceRewriterTests.cs
@@ -373,6 +373,54 @@ public class SwitchCaseBraceRewriterTests
     }
 
     /// <summary>
+    /// Verifies that a single non-terminal statement followed by return makes the switch multi-line
+    /// and adds braces to every non-fall-through section
+    /// </summary>
+    [TestMethod]
+    public void AddsBracesWhenSingleNonTerminalStatementIsFollowedByReturn()
+    {
+        // Arrange
+        const string input = """
+                             class C
+                             {
+                                 bool TryGet(int value, out int result)
+                                 {
+                                     switch (value)
+                                     {
+                                         case 1:
+                                             result = 1;
+
+                                             return true;
+                                         default:
+                                             result = 0;
+
+                                             return false;
+                                     }
+                                 }
+                             }
+                             """;
+
+        // Act
+        var actual = ApplyPhase(input);
+
+        // Assert
+        var tree = CSharpSyntaxTree.ParseText(actual, cancellationToken: TestContext.CancellationToken);
+        var root = tree.GetRoot(TestContext.CancellationToken);
+        var switchStatement = root.DescendantNodes().OfType<SwitchStatementSyntax>().Single();
+
+        foreach (var section in switchStatement.Sections)
+        {
+            Assert.AreEqual(1, section.Statements.Count, "Section should have only the block.");
+            Assert.IsInstanceOfType<BlockSyntax>(section.Statements[0], "Statement should be a block.");
+
+            var block = (BlockSyntax)section.Statements[0];
+
+            Assert.AreEqual(2, block.Statements.Count, "Block should contain the assignment and return.");
+            Assert.IsInstanceOfType<ReturnStatementSyntax>(block.Statements[1], "Return should remain inside the block.");
+        }
+    }
+
+    /// <summary>
     /// Verifies that a trailing comment on the last label (for example "case 1: // note") is
     /// preserved when braces are added
     /// </summary>
--- a/Reihitsu.Formatter.Test/Unit/SwitchCaseBraces/SwitchCaseBraceRewriterTests.cs
+++ b/Reihitsu.Formatter.Test/Unit/SwitchCaseBraces/SwitchCaseBraceRewriterTests.cs
@@ -488,6 +488,100 @@ public class SwitchCaseBraceRewriterTests
     }
 
     /// <summary>
+    /// Verifies that labels with the same name inside separate local functions are not mistaken for
+    /// cross-section goto targets
+    /// </summary>
+    [TestMethod]
+    public void AddBracesIgnoresGotoTargetsInsideLocalFunctions()
+    {
+        // Arrange
+        const string input = """
+                             class C
+                             {
+                                 int M(int value)
+                                 {
+                                     switch (value)
+                                     {
+                                         case 0:
+                                             int F()
+                                             {
+                                                 goto done;
+                                                 done:
+                                                 return 0;
+                                             }
+
+                                             return F();
+                                         case 1:
+                                             int G()
+                                             {
+                                                 goto done;
+                                                 done:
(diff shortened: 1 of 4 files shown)


Reihitsu.Formatter/Pipeline/SwitchCaseBraces/SwitchCaseBraceRewriter.cs
```diff
@@ -76,12 +77,40 @@ private static bool IsMultiLineSection(SwitchSectionSyntax section)
 
         if (statements.Count == 1)
         {
-            return SpansMultipleLines(statements[0]);
+            return SpansMultipleLines(statements[0])
+                   || HasMultiLineBodyEndingInReturnOrThrow(section);
```
thoenissen, line 81: [high] `HasMultiLineBodyEndingInReturnOrThrow` now sends this shape through `AddBraces`, which wraps every section and can make a cross-section label unreachable. For `case 0: goto done; case 1: var result = 1; return result; default: done: return 0;`, the input compiles, but formatted output reports CS0159; refuse brace insertion when it would move a target label into a different block and add this regression case.

Reihitsu.Formatter.Test/Unit/SwitchCaseBraces/SwitchCaseBraceRewriterTests.cs
```diff
@@ -372,6 +372,86 @@ int M(int x)
...
+                                             result = 0;
+
+                                             return false;
+                                     }
+                                 }
+                             }
+                             """;
+
+        // Act
+        var actual = ApplyPhase(input);
```
thoenissen, line 404: [medium] The positive regression calls `ApplyPhase` only once with `Environment.NewLine` and covers only `return`, leaving the new `throw` branch, LF/CRLF behavior, and second-pass idempotency untested. Add full-pipeline assertions under both line endings that format the result again for both terminal forms.

thoenissen: Addressed: added full-pipeline LF/CRLF and second-pass idempotency coverage for statement-plus-return and statement-plus-throw sections (6e558b1c).

Reihitsu.Formatter/Pipeline/SwitchCaseBraces/SwitchCaseBraceRewriter.cs
thoenissen: Addressed: switch brace insertion now stops when an ordinary goto targets a label in another section, with a focused regression test (6e558b1c).

thoenissen: Verified: the original cross-section `goto done` shape is left unbraced, and the focused regression passes.

Reihitsu.Formatter.Test/Unit/SwitchCaseBraces/SwitchCaseBraceRewriterTests.cs
thoenissen: Verified: the return and throw regressions run full-pipeline LF/CRLF and second-pass assertions, and both focused tests pass.

Reihitsu.Formatter/Pipeline/SwitchCaseBraces/SwitchCaseBraceRewriter.cs
```diff
@@ -76,12 +77,79 @@ private static bool IsMultiLineSection(SwitchSectionSyntax section)
...
+    {
+        var labelSections = new Dictionary<string, SwitchSectionSyntax>();
+
+        foreach (var section in node.Sections)
+        {
+            foreach (var labeledStatement in section.DescendantNodes().OfType<LabeledStatementSyntax>())
+            {
+                if (labeledStatement.Ancestors().OfType<SwitchSectionSyntax>().FirstOrDefault() == section)
+                {
+                    labelSections[labeledStatement.Identifier.ValueText] = section;
```
thoenissen, line 130: [high] The flat `ValueText` dictionary conflates labels across executable scopes because `DescendantNodes()` includes local-function bodies: `case 0: int F(){goto done; done:return 0;} return F(); case 1: int G(){goto done; done:return 1;} return G();` compiles, but formatting misclassifies the independent gotos and leaves the multi-line switch unbraced. Restrict collection and matching to the switch's executable scope by skipping nested functions and lambdas, and add this regression.

thoenissen: Addressed: cross-section goto analysis now skips nested local functions and lambdas; focused regressions cover independent labels in both scopes (257a0d71).

Reihitsu.Formatter/Pipeline/SwitchCaseBraces/SwitchCaseBraceRewriter.cs
thoenissen: Still open: an ordinary `goto done` inside a nested switch is excluded because its nearest `SwitchSectionSyntax` is the inner section; formatting then moves the outer target into a sibling block and produces CS0159. Include nested-switch jumps that resolve to labels owned by the outer switch and add this regression.

Reihitsu.Formatter/Pipeline/SwitchCaseBraces/SwitchCaseBraceRewriter.cs
thoenissen: Verified: local-function and lambda bodies are excluded from outer label matching, and both focused regressions now add the switch-section blocks.

Reihitsu.Formatter/Pipeline/SwitchCaseBraces/SwitchCaseBraceRewriter.cs
thoenissen: Addressed: nested-switch jumps now participate when matching labels owned by another outer switch section, with a focused regression (d206a7d8).

thoenissen: Verified at df4b509f: the nested-switch regression passes, the prior counterexample remains unbraced after full CLI formatting, compiles before and after formatting, and is unchanged on a second pass.

PR #568, opened 2026-07-26
Rank ref before partial in the RH7105 modifier order

thoenissen:
## Summary

Give `SyntaxKind.RefKeyword` an explicit rank in `ModifierOrderingUtilities.GetRh7105Rank`, placing it after every other modifier and immediately before `partial`. Adds utility-level ordering tests and analyzer regression tests for ref struct declarations, and updates the RH7105 rule doc to describe the trailing modifier order.

## Why

`ref` had no rank and fell through to the `_ => 100` default, so it sorted behind `partial` (13). C# accepts `ref` only as the last modifier before `partial`, so RH7105 flagged the valid declaration `public ref partial struct S` on its `ref` token, and applying the code fix rewrote it to `public partial ref struct S`, which does not compile.

This is the surviving sibling of the defect class fixed for the `partial`-vs-`readonly/unsafe/async/...` modifiers in #482.

## Linked issues

Closes #487

## Review notes

`ref` now ranks 13 and `partial` moves from 13 to 14; every other rank is unchanged. The ranks are ordinals used only for relative sorting, so the renumbering has no effect beyond placing `ref` between the two groups.

Scope of the behavior change is narrow. `RefKeyword` reaches `GetRh7105Rank` only from `ref struct` declarations — for ref-returning methods, properties, indexers, delegates and partial methods, and for `ref` fields, `ref` is part of the return type or field type and never enters `MemberDeclarationSyntax.Modifiers`. So no ref-returning member is affected. The only reachable change is on declarations carrying both `ref` and `partial`, which is exactly the reported false positive; `readonly ref struct` without `partial` behaved correctly before and still does.

The default rank is left in place for `const` and `fixed`, the two unranked modifiers that still reach the switch. Both have to come last on the declarations that allow them and neither combines with `ref` or `partial`, so rank 100 is already correct for them; the remark on `GetRh7105Rank` names them so the default is not read as unreachable.

One note on test placement. The issue asks for an analyzer test asserting that a misordered `partial ref` is fixed to a compiling order, but that source is not reachable through the analyzer. `public partial ref struct S` does not parse as one declaration: it splits into an `IncompleteMemberSyntax` (`public partial`, modifiers `[public]`) and a `StructDeclarationSyntax` (`ref struct S`, modifiers `[ref]`), reported as CS1585. RH7105 therefore never sees `partial` and `ref` in a single modifier list, and since the parser accepts `ref` in exactly one position, no parseable declaration can present `ref` misordered against another modifier. That reordering is covered in `ModifierOrderingUtilitiesTests`, where the misordered token list can be constructed directly with `SyntaxFactory`.

The analyzer fix test uses a struct with a member because the formatter-backed code fix also collapses an empty type body to a semicolon body, which would otherwise make the test assert an unrelated transform.

## Follow-up work

None.


Reihitsu.Analyzer.Test/Ordering/RH7105DeclarationKeywordsMustFollowOrderAnalyzerTests.cs
```diff
@@ -124,6 +124,95 @@ public readonly partial struct TestStruct
...
+        await Verify(testCode);
+    }
+
+    /// <summary>
+    /// Verifying that a misordered accessibility modifier on a ref partial struct is reported and fixed to a
+    /// compiling order
+    /// </summary>
+    /// <returns>A <see cref="Task"/> representing the asynchronous operation</returns>
+    [TestMethod]
+    public async Task MisorderedModifiersOnRefPartialStructAreReportedAndFixed()
```
thoenissen, line 197: The suite covers only the directive branch of `ModifierOrderingCodeFixProviderBase.ReorderingWouldDropCommentOrDirective` (`MisorderedModifiersWithDirectiveAreReportedWithoutFix`); the comment branch is untested here and in the RH7106 tests. `NormalizeModifierTrivia` rebuilds every modifier token and keeps only the first token's leading trivia, so a regression in that guard silently deletes the comment in `static /* keep */ public int Value { get; set; }`. Add a test asserting that no fix is offered when a comment sits between the modifiers.

---

Reihitsu.Core/ModifierOrderingUtilities.cs
```diff
@@ -127,6 +127,13 @@ public static SyntaxTokenList OrderModifiersForRh7106(SyntaxTokenList modifiers)
     /// </summary>
     /// <param name="modifier">Modifier token</param>
     /// <returns>The ordering rank</returns>
+    /// <remarks>
+    /// The two highest ranks are dictated by the C# grammar rather than by style: <see langword="ref"/> has to
+    /// follow every other modifier and <see langword="partial"/> has to be last, so a ref struct reads
+    /// <c>public readonly ref partial struct</c>. Any modifier that is left unranked falls through to the default
+    /// and therefore sorts behind both of them, which would reorder a valid declaration into one that no longer
```
thoenissen: The remark states that any unranked modifier "would reorder a valid declaration into one that no longer compiles", but two unranked member modifiers already reach this switch and are correct at the default rank: `const` and `fixed` (`unsafe public fixed int B[4];` parses with modifiers `[unsafe, public, fixed]`, and the C# grammar requires both of them last, exactly where rank 100 puts them). Name them so the default is not read as unreachable.

---

thoenissen (commented):
Review of the RH7105 `ref` rank change. The rank placement is correct and the behavior change is as narrow as described: `ref` reaches `GetRh7105Rank` only from type declarations, and the parser accepts `ref` in exactly one position, so the only reachable delta is `ref` versus `partial`. Verified against the Roslyn parser that `public ref partial struct S`, `public readonly ref partial struct S`, `public unsafe ref partial struct S`, `file ref partial struct S` and `public ref partial record struct S` all parse with the full modifier list and no errors, while `public partial ref struct S`, `ref public partial struct S` and `public ref readonly struct S` do not parse into a usable modifier list — so the analyzer-level "misordered `partial ref`" test the issue asks for is indeed unreachable, and covering it in `ModifierOrderingUtilitiesTests` is the right call.

Two findings, both posted inline.

---

Reihitsu.Analyzer.Test/Ordering/RH7105DeclarationKeywordsMustFollowOrderAnalyzerTests.cs
thoenissen: Addressed: added `MisorderedModifiersWithCommentAreReportedWithoutFix`, asserting no fix is offered for `static /* keep */ public int Value { get; set; }` (1725b62).

Verified the test reaches the branch you named rather than the directive one: with `IsCommentOrDirective` temporarily reduced to `trivia.IsDirective`, the new test fails while `MisorderedModifiersWithDirectiveAreReportedWithoutFix` still passes.

Left the RH7106 suite alone — the guard lives on the shared `ModifierOrderingCodeFixProviderBase`, so the comment branch is now covered once for both rules.


---

Reihitsu.Core/ModifierOrderingUtilities.cs
thoenissen: Addressed: the remark now names `const` and `fixed` as the unranked modifiers that are correct at the default rank, and limits the warning to a modifier that must precede a ranked one (0ceb5ec).

Confirmed both parse as you described — `unsafe public fixed int B[4];` gives `[unsafe, public, fixed]` and `public const int C = 1;` gives `[public, const]`, both diagnostic-free, while moving either modifier earlier fails to parse. Neither can combine with `ref` or `partial`, so rank 100 stays correct for them.


---

Reihitsu.Analyzer.Test/Ordering/RH7105DeclarationKeywordsMustFollowOrderAnalyzerTests.cs
thoenissen: Confirmed fixed. `MisorderedModifiersWithCommentAreReportedWithoutFix` hits the `TrailingTrivia` comment branch of `ReorderingWouldDropCommentOrDirective` and asserts no fix is offered; the full RH7105 suite runs green (12/12) at 39f440b. Covering the shared guard once is the right call for RH7106.

---

Reihitsu.Core/ModifierOrderingUtilities.cs
thoenissen: Confirmed fixed. The remark now names `const` and `fixed` as correct at the default rank and narrows the warning to a modifier that must precede a ranked one, which matches the parser behavior.

---

Reihitsu.Analyzer.Test/Naming/RH4115LocalVariableCasingAnalyzerTests.cs
```diff
@@ -576,9 +576,7 @@ public void Load()
     /// <param name="test">Test</param>
     private static void AllowUnsafe(CSharpCodeFixVerifierTest<RH4115LocalVariableCasingAnalyzer, RH4115LocalVariableCasingCodeFixProvider> test)
     {
-        test.SolutionTransforms.Add((solution, projectId) => solution.GetProject(projectId)?.CompilationOptions is CSharpCompilationOptions compilationOptions
-                                                                 ? solution.WithProjectCompilationOptions(projectId, compilationOptions.WithAllowUnsafe(true))
-                                                                 : solution);
+        test.SolutionTransforms.Add(ApplyAllowUnsafeToTestProject);
```
thoenissen, line 578: `CSharpCompilationOptions` was the only consumer of `using Microsoft.CodeAnalysis.CSharp;` in this file, so moving the transform to the base class leaves that directive dead — the project builds clean with it removed. Drop the using.

---

thoenissen (commented):
Re-review of 39f440b. Both prior findings are verified fixed and their threads are resolved: the comment branch of `ReorderingWouldDropCommentOrDirective` is now covered by `MisorderedModifiersWithCommentAreReportedWithoutFix`, and the `GetRh7105Rank` remark names `const` and `fixed` as correct at the default rank. Ran the RH7105 suite (12/12) and the RH4115 suite (13/13) against this head after the `AllowUnsafe` consolidation; both green. The rank change itself is unchanged from the previous pass and still traces clean.

One new finding from the `AllowUnsafe` consolidation, posted inline.

---

Reihitsu.Analyzer.Test/Naming/RH4115LocalVariableCasingAnalyzerTests.cs
thoenissen: Addressed: dropped `using Microsoft.CodeAnalysis.CSharp;` from this file (0eb43ec). Build and the full suite stay green.

`Microsoft.CodeAnalysis.CSharp.Syntax` is kept — `VariableDeclaratorSyntax` still consumes it. Checked the same directive in the RH7105 tests, where it is still live via `SyntaxKind`, so that file is unchanged.


---

thoenissen: Confirmed fixed at 7bda8d3. `Microsoft.CodeAnalysis.CSharp` is gone, `Microsoft.CodeAnalysis.CSharp.Syntax` is still live via `VariableDeclaratorSyntax`, and the RH7105 file's directive is live via the `IsKind` extension — removing it there produces 4 CS errors.

---

Reihitsu.Analyzer.Test/Naming/RH4115LocalVariableCasingAnalyzerTests.cs
```diff
@@ -2,7 +2,6 @@
 using System.Threading.Tasks;
 
 using Microsoft.CodeAnalysis;
```
thoenissen, line 4: `using Microsoft.CodeAnalysis;` is dead here as well — the project builds with 0 errors when it is removed, on this head and on `main`, so it predates the PR. Drop it while this using block is already being cleaned up.

---

thoenissen (commented):
Re-review of 7bda8d3. All three prior findings are verified fixed and their threads are resolved. Rebuilt `Reihitsu.Analyzer.Test` at this head (0 errors) and ran the RH7105 and RH4115 suites (25/25 green). The corrected fixed-buffer example in the `GetRh7105Rank` remark is now valid C#.

One new finding, posted inline: a second dead using directive in the same block that was just cleaned up. It predates the PR — flagged because the file is already open in this diff, not because the PR introduced it.

---

PR #570, opened 2026-07-26
Align directive-wrapped fluent chains under the chain root

thoenissen:
## Summary

`MethodChainAlignmentContributor.ShouldKeepFirstWrappedCallOnContinuationLine` guarded on `HasCommentDirectlyAbove` only. It now mirrors `LineBreakTriviaUtilities.WouldJoinAcrossUnjoinableTrivia`, so a fluent chain whose first wrapped call is held on its continuation line by a preprocessor directive or disabled text aligns under the chain root instead of being flattened to the statement column.

```cs
// before
    return new Builder()
#if DEBUG
    .UseLogging()
#endif
    .UseValidation();

// after
    return new Builder()
#if DEBUG
           .UseLogging()
#endif
           .UseValidation();
```

## Why

Closes the remaining half of #489. The join-corruption half the issue describes (`.Bar();#endif`, CS1040) no longer reproduces — `LineBreakTriviaUtilities.CollapseTokenToSameLine` became directive-aware in the meantime, which shields the comment-only guards in `ChainLineBreakRewriter`. I probed twelve chain/directive shapes through the full pipeline and found no mid-line directive and no non-idempotent pass; `DirectiveAfterTerminatingSemicolonKeepsItsOwnLine` now pins the issue's headline shape so a future change to `CollapseTokenToSameLine` cannot silently re-break it.

What did survive is the de-indent named in the issue title, at the site the issue predicted. The line-break phase correctly refuses to join the first chain dot onto the root line, but the alignment phase did not recognise *why* the dot was still wrapped, skipped the align-under-root path, and let generic block indentation flatten the chain. It was only visible when the chain root does not sit at the statement column, which is why earlier PRs in this area did not surface it.

## Linked issues

Closes #489

## Review notes

- The new predicate is a strict superset of the old one: `ContainsUnjoinableTrivia` covers comments, directives, and disabled text, while `HasCommentDirectlyAbove` additionally required the preceding source line to *look* like a comment. The widening is deliberate and now pinned by `EndOfLineCommentOnRootLineAlignsChainUnderRoot` — a comment in the chain-root token's trailing trivia also aligns the chain under the root.
- Behavior change is limited to chains whose first wrapped call carries unjoinable trivia; `DirectiveAroundMiddleLinkKeepsFirstLinkOnRootLine` pins that a directive around a *middle* link still collapses the first link onto the root line.
- Every case in `MethodChainDirectiveAlignmentTests` runs under both LF and CRLF, asserting the requested ending byte-for-byte plus idempotency.
- `RH5112` already used the same shared predicate, so analyzer and formatter remain in lock-step.

## Follow-up work

**Correction to an earlier claim in this description:** the compiled-out case is *not* purely cosmetic. `reihitsu-format` parses with no preprocessor symbols (`FormatCommandHandler.cs:256`) while the analyzers run with the project's, so in a `DEBUG` build RH5201 reports a diagnostic on formatter-stable output that re-running the formatter cannot converge on.

No guard shipped for it here, because neither candidate works: a `ContainsUnjoinableTrivia` bail-out in RH5201 breaks the existing `VerifyFixPreservesDirectiveInTokenGap` (where all links are active and a stable column does exist), and a disabled-text-only guard can never fire — with `DEBUG` defined there is no disabled text, and without it the compiled-out link is not a chain link at all, so RH5201 already stays silent (`VerifyChainWithDisabledLinkReportsNoDiagnostic` pins this). The real fix is aligning the CLI's parse options with the project's, which is a larger change than this PR.


Reihitsu.Formatter.Test/Regression/Indentation/MethodChainDirectiveAlignmentTests.cs
```diff
@@ -0,0 +1,251 @@
...
+                                            .UseLogging()
+                                #endif
+                                               .UseValidation()
+                                               .Build();
+                                    }
+                                }
+                                """;
+
+        // Act & Assert
+        AssertFormatted(input, expected, false);
```
thoenissen, line 120: The output this test pins is flagged by RH5201, so the "cosmetic only" assessment in the PR body does not hold. `reihitsu-format` parses with default parse options (`Reihitsu.Cli/FormatCommandHandler.cs:256`, no preprocessor symbols), so it leaves the disabled `.UseLogging()` at column 12 while re-aligning the active links to 15. A Debug build defines `DEBUG`, which makes `.UseLogging()` an active chain link, and `RH5201MethodChainsShouldBeAlignedAnalyzer` compares every link column against `chainLinks[0]` with no directive guard at all.

Verified by running RH5201 over this exact expected output with `DEBUG` defined at project level:

```
// /0/Test0.cs(9,16): warning RH5201: Method chains should be aligned.
// /0/Test0.cs(10,16): warning RH5201: Method chains should be aligned.
```

Re-running the CLI cannot converge, so these are permanent warnings on formatter-stable output. Guard RH5201 the way RH5112 already does — bail out via `SyntaxTriviaUtilities.ContainsUnjoinableTrivia` when the link gap carries a directive or disabled text.

---

Reihitsu.Formatter/Pipeline/Indentation/Contributors/MethodChainAlignmentContributor.cs
```diff
@@ -242,7 +247,15 @@ private static bool ShouldKeepFirstWrappedCallOnContinuationLine(SyntaxToken fir
...
-        return ReihitsuFormatterHelpers.HasCommentDirectlyAbove(firstDot);
+        var previousToken = firstDot.GetPreviousToken();
+
+        if (previousToken == default
+            || previousToken.IsKind(SyntaxKind.None))
+        {
+            return SyntaxTriviaUtilities.ContainsUnjoinableTrivia(firstDot.LeadingTrivia);
+        }
+
+        return LineBreakTriviaUtilities.WouldJoinAcrossUnjoinableTrivia(previousToken, firstDot);
```
thoenissen, line 258: `WouldJoinAcrossUnjoinableTrivia` also inspects `anchorToken.TrailingTrivia`, which `HasCommentDirectlyAbove` never did — it required a comment in the moved token's *leading* trivia. An end-of-line comment on the chain-root line therefore takes the exempt path now, and no test covers it. Confirmed behavior change on this input:

```cs
return a // keep this step separate
    .Foo()
    .Bar();
```

main aligns `.Foo()`/`.Bar()` to column 8; this branch aligns them to column 15 under `a`. The new layout is the better one, but add a test with a comment in the anchor token's trailing trivia so the widening is pinned rather than incidental.

---

Reihitsu.Formatter.Test/Regression/Indentation/MethodChainDirectiveAlignmentTests.cs
```diff
@@ -0,0 +1,251 @@
...
+                                    }
+                                }
+                                """;
+
+        // Act & Assert
+        AssertFormatted(input, expected);
+    }
+
+    #endregion // Tests
+}
```
thoenissen, line 248: Issue #489's repro direction names the `.Bar();` + `#endif` shape — the `#endif` sitting after the chain's terminating semicolon. None of the four tests places a directive in that gap (all four put it in a chain-dot gap), and `Regression/LineBreaks/DirectiveJoinTests.cs` has no chain case either, so closing #489 leaves its headline shape unpinned by any test. I confirmed the shape currently formats correctly and idempotently with `DEBUG` defined; add it here so a future change to `CollapseTokenToSameLine` cannot silently re-break it.

---

Reihitsu.Formatter.Test/Regression/Indentation/MethodChainDirectiveAlignmentTests.cs
```diff
@@ -0,0 +1,251 @@
...
+    /// <param name="defineDebug">Whether <c>DEBUG</c> is defined, keeping the conditional branch active</param>
+    /// <param name="cancellationToken">The cancellation token</param>
+    /// <returns>The formatted source text</returns>
+    private static string Format(string input,
+                                 bool defineDebug,
+                                 CancellationToken cancellationToken)
+    {
+        var parseOptions = new CSharpParseOptions(preprocessorSymbols: defineDebug ? ["DEBUG"] : []);
+        var tree = CSharpSyntaxTree.ParseText(input, parseOptions, cancellationToken: cancellationToken);
+        var context = new FormattingContext("\n");
```
thoenissen: `Format` hardcodes `"\n"` for the formatting context and `NormalizeLineEndings` strips CRLF from the expectations, so no case in this class exercises CRLF input. The repo covers this elsewhere by parameterizing `endOfLine` (`AutoPropertyAccessorListCrlfTests`, `SingleLineTypeBraceWithLeadingUsingTests`). I verified CRLF output and idempotency are currently correct for this change; parameterize `endOfLine` so that stays pinned.

---

thoenissen (commented):
Review of the directive-aware chain alignment change. The predicate swap itself is correct and is a strict superset of the old guard; the findings are about analyzer/formatter parity and test coverage. Four findings are inline; two more have no changed-line anchor:

1. **RH5201 code fix and the formatter now target different columns for the same directive-wrapped chain.** `Reihitsu.Analyzer.Test/Formatting/RH5201MethodChainsShouldBeAlignedAnalyzerTests.VerifyFixPreservesDirectiveInTokenGap` asserts the fix produces column 12 for `?.Where`/`?.ToArray` and that the result is diagnostic-clean. Running the formatter over that exact analyzer-clean output moves both links to column 15 (main moved them to 8), so analyzer-clean code is not formatter-stable in either direction. `Reihitsu.Analyzer.Test/Formatter/Formatting/RH5201MethodChainsShouldBeAlignedFormatterTests.cs` contains no `#if` case, which is why the divergence is invisible to CI. Align RH5201's fix with the new chain-root column and add a directive case to the formatter parity tests.

2. **`GetCommentExemptContinuationColumn` and its annotation reason are stale** (`Reihitsu.Formatter/Pipeline/Indentation/Contributors/MethodChainAlignmentContributor.cs:105-118, 140`). The method name, its summary ("a chain that a preceding *comment* keeps wrapped"), and the `"MethodChainCommentExempt"` layout annotation all still say "comment", while the path now also covers preprocessor directives and disabled text. `ShouldKeepFirstWrappedCallOnContinuationLine` got an updated doc; these did not. Rename to match the widened concern so the annotation stays usable for diagnosing layout decisions.

---

Reihitsu.Formatter.Test/Regression/Indentation/MethodChainDirectiveAlignmentTests.cs
thoenissen: Confirmed the RH5201 warnings, but the suggested guard cannot be applied as written, and the narrower variant turns out to be a no-op. Evidence from both directions (9004c85):

- Adding the `ContainsUnjoinableTrivia` bail-out breaks the existing `VerifyFixPreservesDirectiveInTokenGap`, which deliberately pins RH5201 aligning a `#if FEATURE` chain: *"The analyzer did not report a fixable diagnostic."* That coverage is worth keeping — there both links are active, so a single stable column does exist.
- Guarding on disabled text only cannot fire in your repro: with `DEBUG` defined there is no disabled text, and with `DEBUG` undefined the compiled-out link is not a chain link at all, so RH5201 already reports nothing. I verified the latter and added `VerifyChainWithDisabledLinkReportsNoDiagnostic` to pin it.

So no production guard shipped here — it would be dead code. The exposure you found is real and remains open: with `DEBUG` defined, RH5201 reports 1 fixable diagnostic on that output (verified via `GetCodeFixActionsAsync(..., "DEBUG")`).

The root cause is the one you identified in passing: `reihitsu-format` parses with no preprocessor symbols (`FormatCommandHandler.cs:256`) while the analyzer runs with the project's. When both see the same symbols they agree — `DirectiveAboveFirstWrappedCallAlignsChainUnderRoot` is stable and unflagged. Fixing that mismatch is a larger change than this PR; flagging for your call on whether it becomes its own issue.

I have corrected the "cosmetic only" claim in the PR body accordingly.

---

Reihitsu.Formatter/Pipeline/Indentation/Contributors/MethodChainAlignmentContributor.cs
thoenissen: Addressed: added `EndOfLineCommentOnRootLineAlignsChainUnderRoot` in `MethodChainAlignmentFullPipelineTests`, pinning a comment in the chain-root token's trailing trivia aligning the chain under the root rather than at the statement column (9004c85). It runs through `AssertRuleResult`, so LF/CRLF and idempotency are covered too.

---

Reihitsu.Formatter.Test/Regression/Indentation/MethodChainDirectiveAlignmentTests.cs
thoenissen: Addressed: added `DirectiveAfterTerminatingSemicolonKeepsItsOwnLine`, which puts the `#endif` in the gap after the chain's terminating semicolon — the shape named in #489's repro direction — and asserts it keeps its own line and the pass is idempotent (9004c85).

---

Reihitsu.Formatter.Test/Regression/Indentation/MethodChainDirectiveAlignmentTests.cs
thoenissen: Addressed: `Format` now takes `endOfLine`, and `AssertFormatted` loops over LF and CRLF, normalizing input and expectation per ending, asserting the output uses the requested ending byte-for-byte, and re-checking idempotency for each (9004c85). All five cases in the class run under both endings.

---

Reihitsu.Formatter.Test/Regression/Indentation/MethodChainDirectiveAlignmentTests.cs
thoenissen: Your rationale checks out — I reproduced both halves. Adding the `ContainsUnjoinableTrivia` bail-out to `AnalyzeChain` fails `VerifyFixPreservesDirectiveInTokenGap` with exactly "The analyzer did not report a fixable diagnostic" (12 of 13 RH5201 tests still pass), and `VerifyChainWithDisabledLinkReportsNoDiagnostic` passes without any guard. My suggested remedy was wrong; withdraw it.

The finding stays open on the defect, not the fix: in a `DEBUG` build RH5201 still reports on formatter-stable output. Your diagnosis of the root cause — `FormatCommandHandler.cs:256` parsing with no preprocessor symbols while the analyzers use the project's — is the correct one, and out of scope here.

---

Reihitsu.Formatter/Pipeline/Indentation/Contributors/MethodChainAlignmentContributor.cs
thoenissen: Verified fixed. `EndOfLineCommentOnRootLineAlignsChainUnderRoot` pins the comment in the root token's trailing trivia at column 15 under the root, and `AssertRuleResult` runs it under LF and CRLF with the idempotency second pass. Test passes locally.

---

Reihitsu.Formatter.Test/Regression/Indentation/MethodChainDirectiveAlignmentTests.cs
thoenissen: Verified fixed. `DirectiveAfterTerminatingSemicolonKeepsItsOwnLine` places the `#endif` in the gap after the chain's terminating semicolon — the shape #489 names — and asserts it keeps its own line under both endings with idempotency. Test passes locally.

---

Reihitsu.Formatter.Test/Regression/Indentation/MethodChainDirectiveAlignmentTests.cs
thoenissen: Verified fixed. `AssertFormatted` loops both endings, asserts the requested ending byte-for-byte via `AssertUsesLineEnding`, and re-checks idempotency per ending; all five cases pass locally under LF and CRLF. Raised separately: the four line-ending helpers this added are verbatim copies of `FormatterTestsBase`.

---

Reihitsu.Formatter.Test/Regression/Indentation/MethodChainDirectiveAlignmentTests.cs
```diff
@@ -0,0 +1,335 @@
...
+
+        return endOfLine == "\n" ? lineFeedOnly : lineFeedOnly.Replace("\n", endOfLine);
+    }
+
+    /// <summary>
+    /// Verifies that the given text uses the requested end-of-line sequence for every line break
+    /// </summary>
+    /// <param name="text">The formatted text to inspect</param>
+    /// <param name="endOfLine">The end-of-line sequence every line break must use</param>
+    private static void AssertUsesLineEnding(string text, string endOfLine)
```
thoenissen: `_lineEndings` (line 25), `NormalizeLineEndings` (line 67), and `AssertUsesLineEnding` (line 79) are verbatim copies of `Reihitsu.Formatter.Test/Helpers/FormatterTestsBase` — identical bodies, doc comments, and assertion messages — and line 106 inlines its `DescribeLineEnding`. The class stays standalone only because the base's `ApplyRule` hardcodes default parse options. Add an `ApplyRule(string input, string endOfLine, CSharpParseOptions parseOptions)` overload to `FormatterTestsBase` and derive from it, so the line-ending policy lives in one place.

---

Reihitsu.Analyzer.Test/Formatting/RH5201MethodChainsShouldBeAlignedAnalyzerTests.cs
```diff
@@ -641,5 +641,42 @@ void Method()
...
 
+    /// <summary>
+    /// Verifies that a chain carrying a compiled-out link reports no diagnostic. The formatter never
+    /// rewrites disabled text, so the compiled-out link keeps its original column while the active
+    /// links align under the chain root. Demanding a single column across that boundary would flag
+    /// formatter-stable output that re-running the formatter can never converge on (issue #489)
+    /// </summary>
+    /// <returns>A <see cref="Task"/> representing the asynchronous operation</returns>
+    [TestMethod]
+    public async Task VerifyChainWithDisabledLinkReportsNoDiagnostic()
```
thoenissen, line 654: The summary claims this pins a decision about the compiled-out boundary — "demanding a single column across that boundary would flag formatter-stable output" — but `Verify` runs with no preprocessor symbols, so `.UseLogging()` is not a chain link and only the two column-15 links are compared. The assertion cannot fail for the reason the doc gives, and it would still pass under the strictest possible column rule.

The exposure that is actually open runs with `DEBUG` defined, where the same text yields a diagnostic. Either reword the summary to state what it pins (a disabled link is not a chain link) or add the `DEBUG`-defined case documenting the current flagged behavior, so the open parity gap is not read as covered.

---

thoenissen (commented):
Re-review of 909a386. Three of the four inline findings are verified fixed and resolved; the RH5201 exposure stays open with your rationale confirmed (I reproduced the `VerifyFixPreservesDirectiveInTokenGap` failure the suggested guard causes, and withdrew that recommendation on the thread). Two new findings are inline. Two findings from the previous review's summary body are still open and were not addressed:

1. **RH5201's code fix and the formatter still target different columns for the same directive-wrapped chain, with both seeing the same symbols.** Your reply states "when both see the same symbols they agree" — that holds for `DirectiveAboveFirstWrappedCallAlignsChainUnderRoot`, but not for `VerifyFixPreservesDirectiveInTokenGap`, where `#define FEATURE` is in-source so the CLI and the compiler agree on FEATURE. Re-verified on this head: the fix output that test pins (both links at column 12) is not formatter-stable — the formatter moves both to column 15. The mechanism is `RH5201MethodChainsShouldBeAlignedCodeFixProvider.GetReferenceColumn`, which anchors to `links[0]`'s current column and writes whitespace directly instead of delegating to `ReihitsuFormatter.FormatNodeInDocumentAsync` as CLAUDE.md requires. This is independent of the parse-options mismatch and fixable here or in a follow-up. `RH5201MethodChainsShouldBeAlignedFormatterTests` still has no `#if` case, which is why CI does not see it.

2. **`GetCommentExemptContinuationColumn` and its annotation reason are still stale** (`Reihitsu.Formatter/Pipeline/Indentation/Contributors/MethodChainAlignmentContributor.cs:105-118, 140`). The method name, its summary ("a chain that a preceding *comment* keeps wrapped"), and the `"MethodChainCommentExempt"` layout annotation all still say "comment", while the path now also covers directives and disabled text — a gap the newly added `EndOfLineCommentOnRootLineAlignsChainUnderRoot` and the directive tests both exercise.

---

Reihitsu.Formatter.Test/Regression/Indentation/MethodChainDirectiveAlignmentTests.cs
thoenissen: Addressed: added `ApplyRule(input, endOfLine, CSharpParseOptions)` to `FormatterTestsBase`, plus matching `AssertRuleResult` overloads so the ending loop and its assertion messages stay in the base too. `MethodChainDirectiveAlignmentTests` now derives from `FormatterTestsBase` and keeps only a two-line `AssertFormatted` that supplies the parse options; `_lineEndings`, `NormalizeLineEndings`, `AssertUsesLineEnding` and the inlined `DescribeLineEnding` are gone (5877959). The existing no-parse-options overloads delegate to the new ones, so no other fixture changed. All five cases still pass under LF and CRLF.

---

Reihitsu.Analyzer.Test/Formatting/RH5201MethodChainsShouldBeAlignedAnalyzerTests.cs
thoenissen: Addressed by rewording, which is the first of the two options you offered (5877959). The summary now states what the assertion actually pins — parsed without `DEBUG`, the `#if` branch is disabled text, so the compiled-out link is not a chain link and only the two active links are compared — and says explicitly that it does *not* cover the `DEBUG`-defined case, which is still flagged.

I did not add the `DEBUG`-defined case: it would assert the diagnostic we have agreed is the defect, and a passing test asserting that reads as sanctioning it. The gap is recorded in the thread above and in the PR body's follow-up section instead.

---

PR #572, opened 2026-07-26
Correct RH1001 and RH1002 equality rule documentation

thoenissen:
## Summary

Correct the RH1001 and RH1002 rule pages to describe their actual struct equality-performance diagnostics. The pages now list the supported collection and API families, explain the relevant key and comparer behavior, and use matching `IEquatable<T>` struct examples.

## Why

Developers need the help pages to explain when these diagnostics are emitted and how to avoid repeated default value-type equality work in hash-based collections and operations.

## Linked issues

Closes #540

## Review notes

This is documentation-only; analyzer behavior is unchanged. RH1001 checks struct keys and set elements in seven supported collection types. RH1002 checks the equality type selected by its supported operations and exempts calls with an explicit non-null `IEqualityComparer<T>`.

## Follow-up work

None.


thoenissen (commented):
1. **[medium] `documentation/rules/RH1001.md`** — The same analyzer/documentation mismatch remains in the sibling rule: RH1001 only reports struct keys for seven supported collection types, but the page says all key types, describes reference-equality correctness failures, and uses a class example that the analyzer ignores. Update RH1001.md alongside this page so the stale-rule-documentation defect class is closed.

PR #574, opened 2026-07-26
Refuse reorder fixes that would relocate a directive scope

thoenissen:
## Summary

The reorder directive guards no longer decide on the leading trivia of a node's first token. `OrderingMoveSafety.MoveRangeContainsDirectives` inspects the **crossed span** and the **moved span** separately and refuses the move only when a directive in either half would change what it governs:

- a conditional or region directive whose partner lies outside its own half (`SyntaxTriviaUtilities.ContainsUnbalancedConditionalDirectives`, plus the new `ContainsUnbalancedRegionDirectives`);
- any position-sensitive directive — `#pragma`, `#nullable`, `#line` and unknown kinds — which has no partner to balance against.

A pair contained completely inside either half travels intact and is allowed. Both `OrderingDeclarationUtilities` (RH7102/RH7103/RH7109) and `AccessorOrderingUtilities` (RH7107/RH7108) are now thin wrappers over that single shared scan, and all five rule docs describe when the fix is withheld.

## Why

`SyntaxNode.GetLeadingTrivia()` returns the leading trivia of the node's **first** token only. A directive placed between an attribute list (or a modifier) and the declaration keyword attaches to a *later* token, so the guard never saw it and the reorder dragged the enclosed `#if`/`#region` away from its partner — CS1027/CS1028, the exact failure the guard exists to prevent.

```cs
class C
{
    void First()
    {
    }

    [Obsolete]
#if DEBUG
    static void Second()   // RH7103 offered the fix here
    {
    }
#endif
}
```

## Linked issues

Closes #508

## Review notes

- **Why not `ContainsDirectives`.** The first cut used the whole-node flag and withheld the fix whenever *any* directive sat anywhere in the range. That is safe but withholds it for the most common real shape — a balanced `#if DEBUG` inside a crossed member's body, which the move cannot break. The span-pair analysis keeps the issue's shape refused and that one offered.
- **Why the split into two spans.** Only the moved node is relocated, so the two halves must balance independently: a pair straddling the boundary is exactly the case that gets torn apart. A pair enclosing the whole range from outside touches neither half and stays allowed.
- **Region balance is new.** `ContainsPositionSensitiveDirectives` deliberately ignores `#region`, because RH7309's region reorder relocates regions on purpose. A member reorder must not, so `ContainsUnbalancedRegionDirectives` is its counterpart; every directive kind is still covered by exactly one of the three predicates. Both balance predicates share one private nesting walk, parameterized by the opening and closing kind, with the detached `#elif`/`#else` refusal as a conditional-only flag.
- **Verified red-then-green.** The four balanced-pair tests fail against the previous `ContainsDirectives` guard and pass against this one; the directive-blind-spot tests from the original fix fail against `main` and pass here.
- **Docs name the counter-intuitive case.** A `#region` written directly above the first member of the range belongs to that member's leading trivia, so it counts as inside the range and the fix is withheld — the shape `NoCodeFixWhenDirectivesAreInLeadingTrivia` pins. Each rule doc says so explicitly rather than leaving it to contradict the tests.
- `MoveRangeContainsDirectivesReturnsFalseWhenDirectiveSitsOutsideTheMovedRange` pins that the analysed range is still `[target … memberToMove]`.
- The span analysis walks trivia where the old guard read a flag. `OrderingMoveSafety` short-circuits on `root.ContainsDirectives`, but note a single `#region` anywhere in the container already defeats that, so it only spares directive-free code.
- Sibling guards were checked and do not carry the blind spot: `ModifierOrderingCodeFixProviderBase.ReorderingWouldDropCommentOrDirective` already walks every modifier's leading *and* trailing trivia, `UsingDirectiveOrderingSafety` scans the whole using block's `ToFullString()`, and RH7309 uses the span-based predicates. RH7110 ships no code fix.

## Follow-up work

RH7105 and RH7106 also withhold their code fix — `ModifierOrderingCodeFixProviderBase` refuses when a comment or directive sits between modifiers — and their rule docs do not mention it. That guard is untouched here and works differently from the span-pair analysis, so documenting it is left out of this PR.

`#pragma`/`#nullable`/`#line` still refuse in either half, so a self-contained `#pragma warning disable`/`restore` pair inside a body blocks the fix. Allowing it needs disable/restore pairing semantics — per-warning-code lists, `#nullable` targeting annotations vs warnings, `#line default` — that no existing `Reihitsu.Core` predicate models.

`RelocationChangesDirectiveScope` makes three `DescendantTrivia` walks per half. Collapsing them into one combined pass would halve the work, but the single-purpose predicates stay in place for RH7309, so the nesting logic would then exist twice. Left as-is until `FixAll` on large types is actually measured.

`TypeMemberOrderingCodeFixProviderBase.ApplyCodeFixAsync` formats the whole type after a single member move, which widens the fix diff with unrelated edits (visible in the RH7103 expected output: an added `#endregion` comment and a property collapse). Pre-existing and untouched here; narrowing it the way #558 did for RH5109/RH7101 has no reusable member-move transform to delegate to, so it is its own change.

### DIFF
--- a/Reihitsu.Analyzer.CodeFixes/Base/TypeMemberOrderingCodeFixProviderBase.cs
+++ b/Reihitsu.Analyzer.CodeFixes/Base/TypeMemberOrderingCodeFixProviderBase.cs
@@ -59,8 +59,9 @@ public abstract class TypeMemberOrderingCodeFixProviderBase : CodeFixProvider
 
     /// <summary>
     /// Determines whether moving the member before the target member preserves the meaning of the code.
-    /// The base implementation refuses moves that would scramble preprocessor directives sitting anywhere in the
-    /// affected span; derived classes can add further restrictions
+    /// The base implementation refuses moves that would relocate a preprocessor directive away from the code it
+    /// governs; a directive pair contained completely in the moved or the crossed half travels intact and is
+    /// allowed. Derived classes can add further restrictions
     /// </summary>
     /// <param name="typeDeclaration">Containing type declaration</param>
     /// <param name="memberDeclaration">Member declaration to move</param>
--- a/Reihitsu.Analyzer.CodeFixes/Base/TypeMemberOrderingCodeFixProviderBase.cs
+++ b/Reihitsu.Analyzer.CodeFixes/Base/TypeMemberOrderingCodeFixProviderBase.cs
@@ -59,8 +59,8 @@ public abstract class TypeMemberOrderingCodeFixProviderBase : CodeFixProvider
 
     /// <summary>
     /// Determines whether moving the member before the target member preserves the meaning of the code.
-    /// The base implementation refuses moves that would scramble preprocessor directives sitting in the
-    /// affected leading trivia; derived classes can add further restrictions
+    /// The base implementation refuses moves that would scramble preprocessor directives sitting anywhere in the
+    /// affected span; derived classes can add further restrictions
     /// </summary>
     /// <param name="typeDeclaration">Containing type declaration</param>
     /// <param name="memberDeclaration">Member declaration to move</param>
--- a/Reihitsu.Analyzer.Test/Ordering/RH7102ConstantsMustAppearBeforeFieldsAnalyzerTests.cs
+++ b/Reihitsu.Analyzer.Test/Ordering/RH7102ConstantsMustAppearBeforeFieldsAnalyzerTests.cs
@@ -46,7 +46,7 @@ public class RH7102ConstantsMustAppearBeforeFieldsAnalyzerTests : AnalyzerTestsB
 
     /// <summary>
     /// Verifying no code fix is offered when the move would separate a preprocessor directive from its partner,
-    /// leaving the region opened around the target member but closed after the moved member
+    /// with the region opening in the moved field's own leading trivia and closing after it
     /// </summary>
     /// <returns>A <see cref="Task"/> representing the asynchronous operation</returns>
     [TestMethod]
--- a/Reihitsu.Analyzer.Test/Ordering/RH7102ConstantsMustAppearBeforeFieldsAnalyzerTests.cs
+++ b/Reihitsu.Analyzer.Test/Ordering/RH7102ConstantsMustAppearBeforeFieldsAnalyzerTests.cs
@@ -45,7 +45,8 @@ public class RH7102ConstantsMustAppearBeforeFieldsAnalyzerTests : AnalyzerTestsB
     }
 
     /// <summary>
-    /// Verifying no code fix is offered when preprocessor directives sit in the affected leading trivia
+    /// Verifying no code fix is offered when the move would separate a preprocessor directive from its partner,
+    /// leaving the region opened around the target member but closed after the moved member
     /// </summary>
     /// <returns>A <see cref="Task"/> representing the asynchronous operation</returns>
     [TestMethod]
--- a/Reihitsu.Core/AccessorOrderingUtilities.cs
+++ b/Reihitsu.Core/AccessorOrderingUtilities.cs
@@ -79,40 +79,18 @@ public static class AccessorOrderingUtilities
     }
 
     /// <summary>
-    /// Determines whether a preprocessor directive sits anywhere in the span affected by moving an accessor.
-    /// The accessor is moved together with its trivia and jumps over every accessor in between, so directives such as
-    /// <c>#if</c> or <c>#endif</c> would otherwise be dragged to the new position or end up on the other side of the
-    /// moved accessor, splitting a conditional-compilation pair.
-    /// The whole span of each accessor in the range is inspected rather than only the leading trivia of its first
-    /// token, because a directive placed after an attribute list or a modifier attaches to a later token and would
-    /// otherwise stay invisible to the guard
+    /// Determines whether moving an accessor before another accessor would relocate a preprocessor directive away
+    /// from the code it governs, splitting a conditional-compilation pair or scrambling region structure.
+    /// The analysis is shared with the type member guard through
+    /// <see cref="OrderingMoveSafety.MoveRangeContainsDirectives{TNode}"/>
     /// </summary>
     /// <param name="accessorList">Accessor list</param>
     /// <param name="accessorToMove">Accessor to move</param>
     /// <param name="targetAccessor">Target accessor</param>
-    /// <returns><see langword="true"/> if a preprocessor directive sits in the affected span</returns>
+    /// <returns><see langword="true"/> if the move would relocate a preprocessor directive</returns>
     public static bool MoveRangeContainsDirectives(AccessorListSyntax accessorList, AccessorDeclarationSyntax accessorToMove, AccessorDeclarationSyntax targetAccessor)
     {
-        var accessorDeclarations = accessorList.Accessors;
-        var accessorToMoveIndex = accessorDeclarations.IndexOf(accessorToMove);
-        var targetAccessorIndex = accessorDeclarations.IndexOf(targetAccessor);
-
-        if (accessorToMoveIndex < 0
-            || targetAccessorIndex < 0
-            || accessorToMoveIndex <= targetAccessorIndex)
-        {
-            return false;
-        }
-
-        for (var index = targetAccessorIndex; index <= accessorToMoveIndex; index++)
-        {
-            if (accessorDeclarations[index].ContainsDirectives)
-            {
-                return true;
-            }
-        }
-
-        return false;
+        return OrderingMoveSafety.MoveRangeContainsDirectives(accessorList, accessorList.Accessors, accessorToMove, targetAccessor);
     }
 
     #endregion // Methods
--- a/Reihitsu.Core/AccessorOrderingUtilities.cs
+++ b/Reihitsu.Core/AccessorOrderingUtilities.cs
@@ -79,14 +79,18 @@ public static class AccessorOrderingUtilities
     }
 
     /// <summary>
-    /// Determines whether a preprocessor directive sits in the leading trivia affected by moving an accessor.
-    /// The accessor is moved together with its leading trivia, so directives such as <c>#if</c> or <c>#endif</c>
-    /// would otherwise be dragged to the new position, splitting a conditional-compilation pair
+    /// Determines whether a preprocessor directive sits anywhere in the span affected by moving an accessor.
+    /// The accessor is moved together with its trivia and jumps over every accessor in between, so directives such as
+    /// <c>#if</c> or <c>#endif</c> would otherwise be dragged to the new position or end up on the other side of the
+    /// moved accessor, splitting a conditional-compilation pair.
+    /// The whole span of each accessor in the range is inspected rather than only the leading trivia of its first
+    /// token, because a directive placed after an attribute list or a modifier attaches to a later token and would
+    /// otherwise stay invisible to the guard
     /// </summary>
     /// <param name="accessorList">Accessor list</param>
     /// <param name="accessorToMove">Accessor to move</param>
     /// <param name="targetAccessor">Target accessor</param>
-    /// <returns><see langword="true"/> if a preprocessor directive sits in the affected leading trivia</returns>
+    /// <returns><see langword="true"/> if a preprocessor directive sits in the affected span</returns>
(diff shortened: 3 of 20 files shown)


Reihitsu.Core/OrderingDeclarationUtilities.cs
```diff
@@ -253,7 +257,7 @@ public static bool MoveRangeContainsDirectives(TypeDeclarationSyntax typeDeclara
 
         for (var index = targetMemberIndex; index <= memberToMoveIndex; index++)
         {
-            if (members[index].GetLeadingTrivia().Any(trivia => trivia.IsDirective))
+            if (members[index].ContainsDirectives)
```
thoenissen: `ContainsDirectives` is a whole-node flag, so a balanced directive pair fully contained in a crossed member's body now blocks the fix although the move cannot break it — the fix disappears for one of the most common real-world shapes, leaving RH7102/RH7103/RH7109 permanently unfixable:

```cs
class C
{
    void First()
    {
#if DEBUG
        Log();
#endif
    }

    static void Second()
    {
    }
}
```

Before this change RH7103 offered the fix here and produced correct output; now `First.ContainsDirectives` is `true` and the fix is withheld. Narrow the per-node scan to directives that actually travel or change coverage — `SyntaxTriviaUtilities.ContainsUnbalancedConditionalDirectives`, already used by RH7309 for the same hazard, covers the conditional half. The accessor overload in `AccessorOrderingUtilities.cs:109` has the identical regression for RH7107/RH7108, where `#pragma warning disable`/`restore` inside an accessor body is routine.

---

Reihitsu.Core/AccessorOrderingUtilities.cs
```diff
@@ -102,7 +106,7 @@ public static bool MoveRangeContainsDirectives(AccessorListSyntax accessorList,
 
         for (var index = targetAccessorIndex; index <= accessorToMoveIndex; index++)
         {
-            if (accessorDeclarations[index].GetLeadingTrivia().Any(trivia => trivia.IsDirective))
+            if (accessorDeclarations[index].ContainsDirectives)
```
thoenissen: After this change both `MoveRangeContainsDirectives` overloads are identical apart from the list and node types, and this PR exists because they drifted into the same blind spot independently. Extract one generic `SyntaxList<T>` range scan into `Reihitsu.Core` and have both overloads call it, so the next correctness change lands in one place.

---

thoenissen (commented):
The correctness fix itself is sound: `ContainsDirectives` is a strict superset of the previous first-token leading-trivia scan, the widened range still stops at `[targetIndex … moveIndex]`, and the four analyzer tests are meaningful (traced against the old guard, each one previously registered a fix). Regression tests precede the production commit, and issue #508's requirement to apply the change to both overloads is met.

Two inline findings above, plus one that has no changed-line anchor:

1. `documentation/rules/RH7102.md`, `RH7103.md`, `RH7107.md`, `RH7108.md`, `RH7109.md` all advertise **Code Fix ✓** without stating that the fix is withheld when the affected span carries a preprocessor directive. `RH7004.md` and `RH7309.md` document exactly this kind of withholding, and this PR widens it materially — add the corresponding note to the five rule docs.

---

Reihitsu.Core/OrderingDeclarationUtilities.cs
thoenissen: Addressed in df73a2f. The guard no longer looks at the whole node: it inspects the crossed span and the moved span separately and refuses only when a conditional or region directive in either half has its partner outside that half. A pair contained in either half travels intact and is allowed, so your `#if DEBUG` example offers the fix again.

Region balance needed a counterpart to the conditional predicate, since `ContainsPositionSensitiveDirectives` deliberately skips `#region` for RH7309's benefit — added as `SyntaxTriviaUtilities.ContainsUnbalancedRegionDirectives`.

`#pragma`/`#nullable`/`#line` still refuse in either half. They apply from their own position onwards and have no partner to balance against, so proving the coverage survives the move needs disable/restore pairing semantics that no existing Core predicate models; that is recorded as follow-up on the PR rather than guessed at here.

Verified red-then-green: the four new balanced-pair tests fail against the previous `ContainsDirectives` guard and pass against this one.

---

Reihitsu.Core/AccessorOrderingUtilities.cs
thoenissen: Addressed in df73a2f. Both overloads are now thin wrappers over `OrderingMoveSafety.MoveRangeContainsDirectives<TNode>(SyntaxNode root, SyntaxList<TNode> nodes, TNode nodeToMove, TNode targetNode)` in `Reihitsu.Core`, which owns the index resolution, the span computation, and the directive analysis. The narrowing from the other thread landed in that one place and both rules picked it up.

---

Reihitsu.Core/OrderingDeclarationUtilities.cs
thoenissen: Confirmed fixed. `OrderingMoveSafety.MoveRangeContainsDirectives` splits the range into the crossed span and the moved span and refuses only when a conditional or region pair straddles that boundary, so the `#if DEBUG` counterexample offers the fix again — `CodeFixIsOfferedWhenCrossedMemberBodyContainsBalancedConditional` pins the applied output end to end, and the accessor half is covered by `MoveRangeContainsDirectivesReturnsFalseWhenCrossedAccessorBodyContainsBalancedConditional`.

Traced the region predicate against the three positions that decide it — target before `#region`, target after `#region`, target carrying `#endregion` in its leading trivia — and each is classified correctly, including the last one where inserting the moved member at the target's `FullSpan.Start` would drop it inside the region.

The remaining `#pragma`/`#nullable`/`#line` refusal is accepted: they have no partner to balance against, and refusing is the safe direction.

---

Reihitsu.Core/AccessorOrderingUtilities.cs
thoenissen: Confirmed fixed. Both overloads are one-line delegations to `OrderingMoveSafety.MoveRangeContainsDirectives<TNode>`, which owns index resolution, span computation and the directive analysis, and the narrowing landed in that single place.

---

Reihitsu.Analyzer.CodeFixes/Base/TypeMemberOrderingCodeFixProviderBase.cs
```diff
@@ -59,8 +59,8 @@ protected TypeMemberOrderingCodeFixProviderBase(string diagnosticId, string titl
 
     /// <summary>
     /// Determines whether moving the member before the target member preserves the meaning of the code.
-    /// The base implementation refuses moves that would scramble preprocessor directives sitting in the
-    /// affected leading trivia; derived classes can add further restrictions
+    /// The base implementation refuses moves that would scramble preprocessor directives sitting anywhere in the
```
thoenissen: This sentence was written for the `ContainsDirectives` version and is now wrong: the guard no longer refuses a directive sitting *anywhere* in the affected span — a conditional or region pair contained in either half is allowed. `RH7107PropertyAccessorsMustFollowOrderCodeFixProvider.cs:105` and `RH7108EventAccessorsMustFollowOrderCodeFixProvider.cs:64` carry the same stale sentence, while the three `Reihitsu.Core` docs were updated. Reword all three to match `OrderingMoveSafety`: refused when the move would relocate a directive away from the code it governs.

---

thoenissen (commented):
Re-review of the guard rework (791dbd9, df73a2f).

Both inline findings from the previous review are fixed and their threads are resolved. The span-pair analysis is correct: the crossed and moved halves are contiguous and disjoint (`[target.FullSpan.Start, moved.FullSpan.Start)` and `moved.FullSpan`), the moved node lands exactly at `crossedSpan.Start`, so a pair straddling that boundary is precisely the set of relocations that change a directive scope. Traced it against target before `#region`, target after `#region`, target carrying `#endregion` in its leading trivia, `#if` opening in a node before the target, a top-level `#else`, and the `#endif` that attaches to the accessor list's closing brace — each is classified correctly, and the previously passing RH7103 leading-trivia and region tests still refuse.

One new finding inline. One item from the previous review is still open:

1. `documentation/rules/RH7102.md`, `RH7103.md`, `RH7107.md`, `RH7108.md`, `RH7109.md` still advertise **Code Fix ✓** without stating that the fix is withheld when the move would relocate a directive scope. `RH7004.md` and `RH7309.md` document this kind of withholding; the narrowing makes the condition easier to state, not less necessary.

---

Reihitsu.Analyzer.CodeFixes/Base/TypeMemberOrderingCodeFixProviderBase.cs
thoenissen: Addressed in f329596. All three now say the move is refused when it would relocate a directive away from the code it governs, with a pair contained in the moved or the crossed half allowed.

Grepping the phrasing turned up five more instances of the same stale rule in the ordering test summaries (RH7102/RH7103/RH7109 `NoCodeFixWhenDirectivesAreInLeadingTrivia`, RH7107/RH7108 `NoCodeFixWhenDirectivesAreInAccessorLeadingTrivia`). Those are corrected in 15f8485 to state the actual reason — the move would separate the directive from its partner. No occurrence of the old wording remains in the repository.

---

thoenissen: Confirmed fixed. All three now describe the actual rule, and the shared-helper refactor in e1d6121 is behaviour-preserving — `refuseDetachedBranches: true` reproduces the conditional branch exactly, `false` short-circuits it for regions, and `ContainsUnbalancedRegionDirectivesReturnsFalseForDetachedElseDirective` pins that.

One inaccuracy in the follow-up commit 15f8485 is raised separately: the RH7102 and RH7109 summaries describe RH7103's shape.

---

Reihitsu.Analyzer.Test/Ordering/RH7102ConstantsMustAppearBeforeFieldsAnalyzerTests.cs
```diff
@@ -45,7 +45,8 @@ public class TestClass
     }
 
     /// <summary>
-    /// Verifying no code fix is offered when preprocessor directives sit in the affected leading trivia
+    /// Verifying no code fix is offered when the move would separate a preprocessor directive from its partner,
+    /// leaving the region opened around the target member but closed after the moved member
```
thoenissen: This summary describes RH7103's shape, not this test's. Here `#region Constants` opens immediately before the moved member `MaxValue` and the target `_value` sits outside the region entirely, so the guard refuses because the region opens inside the *moved* half with its `#endregion` outside it. `RH7109ReadonlyElementsMustAppearBeforeNonReadonlyElementsAnalyzerTests.cs:106` has the same test shape and the same wrong summary. Only RH7103, where `#region Fields` precedes the target member, matches the sentence as written.

---

thoenissen (commented):
Re-review of f329596, e1d6121, 15f8485, e8dcdbc and the main merge.

The stale-documentation finding is fixed and its thread is resolved. The `ContainsUnbalancedDirectives` extraction is behaviour-preserving: with `refuseDetachedBranches: true` the conditional predicate reproduces its previous branch order exactly, with `false` the detached-branch arm is short-circuited, the null-root refusal is preserved in the shared helper, and the new detached-`#else` test pins the region side. `OrderingMoveSafety` is unchanged apart from the corrected short-circuit comment, so the span-pair analysis verified in the previous round still stands.

One new finding inline. One item from the previous review is still open:

1. `documentation/rules/RH7102.md`, `RH7103.md`, `RH7107.md`, `RH7108.md`, `RH7109.md` still advertise **Code Fix ✓** without stating that the fix is withheld when the move would relocate a directive scope. `RH7004.md` and `RH7309.md` document this kind of withholding, and the XML docs corrected in f329596 now carry a precise sentence the rule docs could reuse.

---

Reihitsu.Analyzer.Test/Ordering/RH7102ConstantsMustAppearBeforeFieldsAnalyzerTests.cs
thoenissen: Correct, and fixed in 2811440. Both RH7102 and RH7109 now read "with the region opening in the moved field's own leading trivia and closing after it". Re-traced the other three against their own sources: RH7103 keeps the original sentence because `#region Fields` does precede the target there, and RH7107/RH7108 describe the conditional opening around the target accessor and closing before the moved one, which matches their shapes.

---

PR #575, opened 2026-07-26
Preserve documentation boundaries in blank-line cleanup

thoenissen:
## Summary

Preserve the mandatory line break after multi-line documentation comments while removing genuine blank-line runs in one pass, including trailing whitespace and repeated documentation blocks. Centralize embedded line-break detection, preserve intervening comments and directives, and align RH8302 analyzer, code-fix, formatter parity, and rule documentation with `/** */` comments.

## Why

Multi-line documentation comments expose their terminating line break as separate trivia, so blank-line cleanup could join declarations, defer cleanup to a second pass, or leave earlier gaps between stacked comments. RH8302 previously recognized only `///` and removed only one line per action, leaving formatter parity and fix convergence incomplete.

## Linked issues

Closes #526

## Review notes

RH8302 now reports the complete contiguous blank-line run after either documentation-comment form, so one code-fix or Fix All iteration converges. Formatter cleanup processes every documentation trivia in a token, counts explicit line breaks relative to embedded terminators, scans past trailing spaces and tabs, and retains ordinary comments or preprocessor directives in the affected gap.

Review follow-up excludes synthetic or unterminated EOF lines from RH8302 spans, covers LF and CRLF convergence, and retains the cheap trivia-kind guard before materializing trivia text.

## Follow-up work

None.

### DIFF
--- a/Reihitsu.Analyzer/Rules/Documentation/RH8302ElementDocumentationHeadersMustNotBeFollowedByBlankLineAnalyzer.cs
+++ b/Reihitsu.Analyzer/Rules/Documentation/RH8302ElementDocumentationHeadersMustNotBeFollowedByBlankLineAnalyzer.cs
@@ -1,4 +1,5 @@
 ﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Text;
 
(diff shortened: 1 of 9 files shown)


Reihitsu.Analyzer/Rules/Documentation/RH8302ElementDocumentationHeadersMustNotBeFollowedByBlankLineAnalyzer.cs
```diff
@@ -38,6 +39,89 @@ public RH8302ElementDocumentationHeadersMustNotBeFollowedByBlankLineAnalyzer()
...
+    /// <param name="firstLineIndex">Index of the first possible blank line</param>
+    /// <param name="blankLineRun">Span of the complete blank-line run</param>
+    /// <returns><see langword="true"/> if a blank-line run was found; otherwise, <see langword="false"/></returns>
+    private static bool TryGetBlankLineRun(SourceText sourceText,
+                                           int firstLineIndex,
+                                           out TextSpan blankLineRun)
+    {
+        blankLineRun = default;
+
+        if (firstLineIndex >= sourceText.Lines.Count
```
thoenissen, line 55: This accepts the synthetic or unterminated final SourceText line as a removable blank-line run. With `/** <summary>Dangling.</summary> */\n`, RH8302 is reported at a zero-length EOF span and the registered fix replaces empty text with empty text, so require the candidate line to contain a real line break and add LF/CRLF convergence coverage.

Reihitsu.Formatter/Pipeline/BlankLines/BlankLineCollapser.cs
```diff
@@ -53,7 +52,7 @@ private static SyntaxTriviaList CollapseBlankLinesInTrivia(SyntaxTriviaList triv
         {
             currentLine.Add(triviaItem);
 
-            if (triviaItem.IsKind(SyntaxKind.EndOfLineTrivia) || EndsWithEmbeddedLineBreak(triviaItem))
+            if (triviaItem.IsKind(SyntaxKind.EndOfLineTrivia) || BlankLineTriviaUtilities.EndsWithLineBreak(triviaItem))
```
thoenissen, line 55: This new shared call reaches `ToFullString()` for every trivia item processed here, including ubiquitous whitespace, whereas the previous implementation filtered to directives, disabled text, and documentation comments first. Restore that cheap kind guard before materializing trivia text to avoid a formatter-wide allocation regression.

thoenissen (commented):
1. The #526 regression tests and the production fix first appear together in commit 05b3553b; the required failing regression-test commit does not precede the implementation. Split or reorder the history so the regression test demonstrably fails before the formatter fix.

Reihitsu.Analyzer/Rules/Documentation/RH8302ElementDocumentationHeadersMustNotBeFollowedByBlankLineAnalyzer.cs
thoenissen: Addressed in 07d06abc: RH8302 now rejects blank-line candidates without an actual line break, with LF and CRLF EOF regression coverage.

Reihitsu.Formatter/Pipeline/BlankLines/BlankLineCollapser.cs
thoenissen: Addressed in 07d06abc: the shared helper restores the directive, disabled-text, and documentation trivia guard before calling ToFullString().

Reihitsu.Analyzer/Rules/Documentation/RH8302ElementDocumentationHeadersMustNotBeFollowedByBlankLineAnalyzer.cs
thoenissen: Verified: EOF candidates without a physical line break are rejected, and LF/CRLF regression tests cover the multi-line documentation case.

Reihitsu.Formatter/Pipeline/BlankLines/BlankLineCollapser.cs
thoenissen: Verified: ToFullString() is again gated to directives, disabled text, and documentation trivia.

PR #578, opened 2026-07-26
Consolidate the Core safety predicates and fix the guards that discarded documentation comments

thoenissen:
## Summary

Collapses the duplicated safety-predicate families onto single `Reihitsu.Core` helpers, deletes the stale private copies in `Reihitsu.Formatter`, `Reihitsu.Analyzer` and `Reihitsu.Analyzer.CodeFixes`, and corrects the guards whose comment set or scope did not match the rewrite they protect.

**Consolidation**

- `FormattingSafetyUtilities` is deleted; its members move to `SyntaxNodeUtilities` (`IsSingleLineSpan`, `AreAllSingleLine`), with the three single-line checks sharing one line-span primitive.
- `SyntaxTriviaUtilities.IsRegionDirective` becomes the canonical region-directive predicate, replacing copies in `RegionDirectiveBlankLineUtilities`, `RegionDirectiveUtilities`, `BlankLineRegionDirectiveRewriter`, `LayoutComputer`, `NestedRegionRemovalStep`, `IndentationRewriter`, the RH5204, RH7303 and RH7310 analyzers and the RH7304 code fix.
- `SyntaxNodeUtilities.IsComment` is dropped for `SyntaxTriviaUtilities.IsCommentTrivia`; the private comment copies in `UsingDirectiveOrderingUtilities` and `RH7310EmptyRegionsShouldBeRemovedAnalyzer` and the inline region copies in `RegionDirectiveUtilities` are removed too.
- `SyntaxNodeUtilities` now exposes three node-scoped comment-or-directive predicates that differ only in scope, each documented against the others: `ContainsCommentOrDirective` (full trivia), `InteriorContainsCommentOrDirective` (the node's own span), and `GroupInteriorContainsCommentOrDirective` (a sibling group plus the gaps between them). `ContainsNonDocumentationCommentOrDirective` is the deliberately narrow one.

**Behaviour changes**

This is not a behaviour-preserving refactor. Consolidating the predicates exposed guards whose comment set or scope was wrong for what they protect, each verified by running `reihitsu-format` or the code fix in question:

- The auto-property accessor collapse **deleted** a documentation comment (`PropertyLayoutLineBreakRewriter`, mirrored by the RH5408 analyzer and fix); the attribute-list merge deleted one sitting between two lists (`AttributeTargetFormattingRewriter`, mirrored by `TargetAttributeListShapeCodeFixProviderBase`).
- RH5307 reported diagnostics whose fix the formatter then refused, leaving a diagnostic the user could not clear. `LineBreakListRewriter` kept an inline copy of the same narrow predicate, the formatter twin of that guard.
- The RH3002, RH3004, RH3005, RH3102 and RH5109 fixes discarded a comment when removing parentheses, converting an anonymous method, swapping a comparison, removing an empty statement, or rebuilding a parameter list. RH3005 and RH5109 rebuild their result from raw text or from a reparse instead of delegating to the formatter, so anything outside the copied spans was lost; both now refuse. RH3004 still offers its fix and falls back to preserving the block, and additionally refuses when trivia sits between the delegate keyword and the parameter list, or between the parameter list and the block.
- The attribute placement fix rewrites the closing bracket's trailing trivia and clears the member token's leading trivia, so a comment between them was destroyed under either placement mode; the gap guard now applies to both, in the fix and in `TargetAttributePlacementAnalyzerBase` alike, so the fourteen separate-line rules no longer report a diagnostic the fix refuses.
- RH5111 changes deliberately: a documented member, or one preceded by a region directive, now **does** get the fix, because the join never reaches the trivia in front of the declaration. RH5101 and RH5102 keep their existing behaviour, now via the shared predicate.

RH5306 is unchanged and keeps the narrow predicate: its fix delegates to the formatter, and the added test pins that a documented list pattern keeps its comment and converges.

## Why

Closes the round-2 review finding that the round-1 consolidation recreated duplication inside Core and left byte-equivalent private copies alive outside it. Each copy was a place where a comment- or directive-guard could drift from the shared predicate — and the drift had already happened: several guards omitted documentation comments while the formatter's own join guard counts them, which is what produced the source loss above.

## Linked issues

Closes #420

## Review notes

Scope is the subtle part. `ContainsCommentOrDirective` walks a node's **full** trivia, so for a node that begins a member declaration — an attribute list — it also counts that member's documentation comment. Using it there made a documented class stop being reshaped while an identical undocumented one still was, and cost it its code fix. Split and placement rewrites therefore use the interior-scoped predicate, while the merge rewrite uses the group-scoped one so it still refuses when trivia sits between the lists it folds together. Core tests pin all three scopes explicitly.

Regression tests were written before each fix and cover LF, CRLF and second-pass idempotency for the formatter, and registration or convergence for RH3002, RH3004, RH3005, RH3102, RH5109, RH5111, RH5306, RH5307, RH5408, RH5505 and RH5506.

Issue #420 asks for the RH5101/RH5102/RH5111 guards to keep behaviour unchanged. RH5111 deliberately deviates, and that deviation is **signed off**: guard behaviour may change where the audit found a real problem, and RH5111 is one. Narrowing it to the member's own span is what stops a documented member from losing a fix an undocumented one still gets — holding the old behaviour would have preserved the defect rather than the guard's intent. RH5101 and RH5102 are unchanged, as the issue asks.

Two pre-existing defects found while verifying are deliberately **not** fixed here, as neither involves these predicates: an interior `/** */` in an attribute list is moved onto its own line by the documentation-comment phase and then oscillates with the blank-line phase, and a trailing `///` on a member line is truncated, destroying its text.

## Follow-up work

- `RH5308ConditionalExpressionsShouldBeFormattedCorrectlyAnalyzer`, `RH5405BracesMustNotBeOmittedAnalyzer`, `RH5406BracesMustNotBeOmittedFromMultiLineChildStatementsAnalyzer:72-74` and `SwitchCaseBraceRewriter.SpansMultipleLines:198-203` still compare line positions inline instead of calling `SyntaxNodeUtilities.IsSingleLine` / `IsSingleLineSpan`.
- `BlankLineEditor.cs:204` still matches region directives by `SyntaxKind` inline rather than through `SyntaxTriviaUtilities.IsRegionDirective`.
- `TargetAttributeListShapeAnalyzerBase.cs:77` has no comment guard at all, a pre-existing analyzer/fix parity gap that this PR does not widen.
- `FieldDeclarationSplitTransform.cs:83` uses a documentation-comment-blind comment set: a documented multi-variable field split is mis-indented on the first pass and self-corrects on the second. Identical on `main`.
- The two pre-existing documentation-comment formatting defects noted above.


Reihitsu.Core/SyntaxNodeUtilities.cs
```diff
@@ -34,16 +35,26 @@ public static bool HasCommentsOrDirectives(SyntaxNode node)
...
+    /// comments, matching <see cref="SyntaxTriviaUtilities.ContainsUnjoinableTrivia"/>, which the formatter
+    /// consults before collapsing a multi-line construct. Analyzers and code fixes that predict a formatter
+    /// join at registration time use this predicate so they never offer an action the formatter then refuses
     /// </summary>
-    /// <param name="trivia">Trivia</param>
-    /// <returns><see langword="true"/> if the trivia is a comment; otherwise <see langword="false"/></returns>
-    public static bool IsComment(SyntaxTrivia trivia)
+    /// <param name="node">Node</param>
+    /// <returns><see langword="true"/> if the node contains join-refusing trivia; otherwise <see langword="false"/></returns>
+    public static bool ContainsJoinRefusingTrivia(SyntaxNode node)
```
thoenissen: `ContainsJoinRefusingTrivia` re-implements `IsCommentOrDirective` from the same file (line 65): its loop condition `trivia.IsDirective || SyntaxTriviaUtilities.IsCommentTrivia(trivia)` is that predicate verbatim, just with the operands swapped. A consolidation PR should not ship a fresh copy of a predicate that lives 20 lines below it. Replace the body with `return node.DescendantTrivia(descendIntoTrivia: true).Any(IsCommentOrDirective);`, which also makes it the node-scoped sibling of the existing `SpanContainsCommentOrDirective`.

---

Reihitsu.Core/SyntaxNodeUtilities.cs
```diff
@@ -34,16 +35,26 @@ public static bool HasCommentsOrDirectives(SyntaxNode node)
     }
 
     /// <summary>
-    /// Determines whether the given trivia is a comment
+    /// Determines whether a node contains trivia that makes the formatter refuse to join its lines. The
+    /// comment set is wider than <see cref="HasCommentsOrDirectives"/>: it also covers documentation
+    /// comments, matching <see cref="SyntaxTriviaUtilities.ContainsUnjoinableTrivia"/>, which the formatter
```
thoenissen: The doc states this predicate matches `SyntaxTriviaUtilities.ContainsUnjoinableTrivia`, but the two trivia sets differ: `ContainsUnjoinableTrivia` blocks on `IsDirectiveOrDisabledTextTrivia` (directives **and** `DisabledTextTrivia`), while this one checks `IsDirective` only. State the actual set instead of claiming equivalence — an inaccurate equivalence note is the seed of the drift this PR removes. Also align the vocabulary: `ContainsJoinRefusingTrivia` and `ContainsUnjoinableTrivia` are two names for one concept living in the same assembly.

---

thoenissen (commented):
Every copy issue #420 enumerates is gone and the replacements are semantically identical to what they replace, including the deliberate `ContainsJoinRefusingTrivia` / `HasCommentsOrDirectives` doc-comment split. Findings below; two are inline, the rest have no changed-line anchor.

1. `Reihitsu.Formatter/Pipeline/LineBreaks/LineBreakListRewriter.cs:203-204` keeps an inline copy of `HasCommentsOrDirectives` (`trivia.IsDirective || SingleLineCommentTrivia || MultiLineCommentTrivia`). It sits in `CanSafelyCollapseBracketedArguments`, the formatter twin of `RH5307IndexerBracketedArgumentsShouldBeSingleLinedAnalyzer.CanSafelyCollapseToSingleLine` — which this PR redirected to `SyntaxNodeUtilities.HasCommentsOrDirectives`. The analyzer and its code fix now share the Core predicate while the formatter phase they must agree with keeps a private copy, which is exactly the analyzer/formatter drift this consolidation exists to close. Redirect it to `SyntaxNodeUtilities.HasCommentsOrDirectives`.

2. Two byte-equivalent copies of `SyntaxTriviaUtilities.IsCommentTrivia` survive and are named neither in the issue nor in the Follow-up section: `Reihitsu.Core/UsingDirectiveOrderingUtilities.cs:317-323` (`HasCommentTrivia`) and `Reihitsu.Analyzer/Rules/Organization/RH7310EmptyRegionsShouldBeRemovedAnalyzer.cs:83-86` (inlined in `ContainsComment`). The first is Core-internal duplication of the four-kind comment predicate — the defect class issue #420 names first. Redirect both to `SyntaxTriviaUtilities.IsCommentTrivia`.

3. The Follow-up section understates the surviving single-line-comparison family. Besides RH5308 and RH5405: `Reihitsu.Analyzer/Rules/Layout/RH5408SimpleAutoPropertiesShouldBeSingleLinedAnalyzer.cs:181-183` still inlines the comparison in a file this PR otherwise moved onto `SyntaxNodeUtilities.IsSingleLineSpan`; `Reihitsu.Formatter/Pipeline/SwitchCaseBraces/SwitchCaseBraceRewriter.cs:198-203` (`SpansMultipleLines`) is a named private copy of the same kind the issue enumerates; `Reihitsu.Analyzer/Rules/Layout/RH5406BracesMustNotBeOmittedFromMultiLineChildStatementsAnalyzer.cs:72-74` inlines it too. Convert RH5408:181-183 to `SyntaxNodeUtilities.IsSingleLineSpan(propertyDeclaration.SyntaxTree, relevantSpan)` and correct the Follow-up list to the full set.

---

Reihitsu.Core/SyntaxNodeUtilities.cs
thoenissen: Addressed: the body is now `node.DescendantTrivia(descendIntoTrivia: true).Any(IsCommentOrDirective)` instead of re-implementing that predicate, and the method moved next to `SpanContainsCommentOrDirective` so the node-scoped and span-scoped forms sit together (8552ffe).

---

Reihitsu.Core/SyntaxNodeUtilities.cs
thoenissen: Addressed: the doc no longer claims equivalence. It now states the actual set — comments including documentation comments, plus directives — and records that the formatter's guard additionally blocks on disabled text, so the two are explicitly not identical.

On the vocabulary: rather than sharing a name with `ContainsUnjoinableTrivia` while the sets differ, the predicate is renamed `ContainsCommentOrDirective`, aligning it with the `IsCommentOrDirective` / `SpanContainsCommentOrDirective` family in the same file — the sibling relationship you pointed at in the other thread. The name now describes the actual trivia set, and the doc carries the join-prediction purpose (8552ffe).

---

Reihitsu.Core/SyntaxNodeUtilities.cs
thoenissen: Verified fixed: the body is `node.DescendantTrivia(descendIntoTrivia: true).Any(IsCommentOrDirective)` and the method sits next to `SpanContainsCommentOrDirective`. No copy of the predicate remains.

---

Reihitsu.Core/SyntaxNodeUtilities.cs
thoenissen: Verified fixed: the doc states the actual trivia set and records that the formatter's guard additionally blocks on disabled text, instead of claiming equivalence.

The rename raises a separate problem, filed as a new finding: `ContainsCommentOrDirective(SyntaxNode)` and `HasCommentsOrDirectives(SyntaxNode)` now read as synonyms in the same class with different trivia sets.

---

Reihitsu.Formatter/Pipeline/LineBreaks/PropertyLayoutLineBreakRewriter.cs
```diff
@@ -169,7 +140,7 @@ private static SyntaxToken GetSingleLineSignatureStartToken(PropertyDeclarationS
     /// </remarks>
     private static bool CanCollapseAutoPropertyToSingleLine(PropertyDeclarationSyntax node)
     {
-        if (node?.AccessorList == null || HasCommentsOrDirectives(node.AccessorList))
+        if (node?.AccessorList == null || SyntaxNodeUtilities.HasCommentsOrDirectives(node.AccessorList))
```
thoenissen: This guard uses the narrow `HasCommentsOrDirectives`, which does not block documentation comments, so the collapse silently deletes one. Verified by running `reihitsu-format` over this file:

```csharp
internal class TestClass
{
    public int Value
    {
        /// <summary>note</summary>
        get;
        set;
    }
}
```

Output: `public int Value { get; set; }` — the documentation comment is gone. This is source loss in the shipped CLI, not just an IDE code action; RH5408's analyzer (`:103`) and code fix (`:62`) carry the same guard and delete it the same way. Switch all three to `SyntaxNodeUtilities.ContainsCommentOrDirective`.

---

Reihitsu.Analyzer/Rules/Layout/RH5307IndexerBracketedArgumentsShouldBeSingleLinedAnalyzer.cs
```diff
@@ -55,8 +55,8 @@ private static bool IsIndexerArgumentList(BracketedArgumentListSyntax bracketedA
     /// <returns><see langword="true"/> if collapsing is safe; otherwise, <see langword="false"/></returns>
     private static bool CanSafelyCollapseToSingleLine(BracketedArgumentListSyntax bracketedArgumentList)
     {
-        return FormattingSafetyUtilities.HasCommentsOrDirectives(bracketedArgumentList) == false
-               && FormattingSafetyUtilities.AreAllSingleLine(bracketedArgumentList.Arguments);
+        return SyntaxNodeUtilities.HasCommentsOrDirectives(bracketedArgumentList) == false
```
thoenissen: The narrow `HasCommentsOrDirectives` lets a documentation comment through, and the resulting fix does not converge. Verified: applying the fix to

```csharp
return values[
    /// <summary>note</summary>
    0];
```

leaves the list multi-line, inserts a blank line after `[`, and the diagnostic reports again on the fix output — a permanent RH5307 the user cannot clear. The code fix at `RH5307...CodeFixProvider.cs:56` has the same guard. Use `SyntaxNodeUtilities.ContainsCommentOrDirective` on both sides.

---

Reihitsu.Core/SyntaxNodeUtilities.cs
```diff
@@ -77,7 +70,24 @@ public static bool GapContainsComment(SyntaxToken firstToken, SyntaxToken second
...
+    /// of <see cref="IsCommentOrDirective"/>, next to the span-scoped <see cref="SpanContainsCommentOrDirective"/>.
+    /// Unlike <see cref="HasCommentsOrDirectives"/> the comment set includes documentation comments, which is what
+    /// analyzers and code fixes need when they predict at registration time whether the formatter will refuse to
+    /// join lines: the formatter's join guard <see cref="SyntaxTriviaUtilities.ContainsUnjoinableTrivia"/> counts
+    /// documentation comments too, so a narrower guard registers actions the formatter then refuses. The two sets
+    /// are not identical — the formatter's guard additionally blocks on disabled text
+    /// </summary>
+    /// <param name="node">Node</param>
+    /// <returns><see langword="true"/> if a comment or directive is present; otherwise <see langword="false"/></returns>
+    public static bool ContainsCommentOrDirective(SyntaxNode node)
```
thoenissen, line 88: The rename creates a name collision that the previous name did not have: `ContainsCommentOrDirective(SyntaxNode)` (3 call sites, includes documentation comments) and `HasCommentsOrDirectives(SyntaxNode)` at `:27` (21 call sites, excludes them) are synonyms in the same class with the same signature and different trivia sets. IntelliSense offers both with nothing to choose by, and the three high findings in this review are exactly that wrong pick already made. Put the distinction in the narrow one's name — for example `ContainsNonDocumentationCommentOrDirective` — so the wider predicate is the unsurprising default.

---

thoenissen (commented):
Re-review of `354cbd2`. Both inline findings from the previous review are verified fixed and their threads are resolved. Three of the five prior findings are still open, and tracing the narrow/wide predicate split this PR formalises turned up three confirmed trivia-loss and convergence defects on the narrow side.

**New findings without a changed-line anchor**

1. `Reihitsu.Analyzer.CodeFixes/Rules/Layout/RH5109ParametersMustBeOnSameLineOrSeparateLinesCodeFixProvider.cs:99` guards with the narrow `HasCommentsOrDirectives`, then rebuilds the parameter list from raw `parameter.Span` text at `:51`/`:61`. A documentation comment between parameters passes the guard and is outside every `parameter.Span`, so it is silently deleted. Verified — applying the fix to `void Method(int first, int second,` / `/// <summary>note</summary>` / `int third)` yields a parameter list with the comment gone. The comment at `:64-65` claiming "the registration rejects parameter lists that carry comments or directives" is false for documentation comments. Switch the guard to `SyntaxNodeUtilities.ContainsCommentOrDirective`.

**Prior findings still open**

2. `Reihitsu.Formatter/Pipeline/LineBreaks/LineBreakListRewriter.cs:203-204` still holds an inline copy of `HasCommentsOrDirectives` in `CanSafelyCollapseBracketedArguments`, the formatter twin of the RH5307 guard this PR redirected to Core. Unchanged since the last review.

3. Two byte-equivalent copies of `SyntaxTriviaUtilities.IsCommentTrivia` still survive: `Reihitsu.Core/UsingDirectiveOrderingUtilities.cs:317-323` (`HasCommentTrivia`, Core-internal — the defect class issue #420 names first) and `Reihitsu.Analyzer/Rules/Organization/RH7310EmptyRegionsShouldBeRemovedAnalyzer.cs:83-86`. Unchanged since the last review.

4. The Follow-up section still lists only RH5308 and RH5405 for the surviving single-line comparison. `Reihitsu.Analyzer/Rules/Layout/RH5408SimpleAutoPropertiesShouldBeSingleLinedAnalyzer.cs:181-183` (a file this PR otherwise moved onto `SyntaxNodeUtilities.IsSingleLineSpan`), `Reihitsu.Formatter/Pipeline/SwitchCaseBraces/SwitchCaseBraceRewriter.cs:198-203` (a named private copy of the kind the issue enumerates) and `Reihitsu.Analyzer/Rules/Layout/RH5406BracesMustNotBeOmittedFromMultiLineChildStatementsAnalyzer.cs:72-74` all still carry it. Unchanged since the last review.

---

Reihitsu.Formatter/Pipeline/LineBreaks/PropertyLayoutLineBreakRewriter.cs
thoenissen: Addressed: the accessor-list guard, the RH5408 analyzer (`:103`) and its code fix (`:62`) now use `SyntaxNodeUtilities.ContainsCommentOrDirective`, so the collapse is refused and the documentation comment survives. Regression tests cover LF, CRLF and second-pass idempotency for the formatter, plus the analyzer and the code fix.

The same defect class was closed elsewhere in the same pass: the attribute-list merge deleted a comment between two lists, and the RH3002, RH3004 and RH3102 fixes discarded one when rewriting. RH5306 carried it too and was not in the review.

---

Reihitsu.Analyzer/Rules/Layout/RH5307IndexerBracketedArgumentsShouldBeSingleLinedAnalyzer.cs
thoenissen: Addressed: the RH5307 analyzer (`:58`) and its code fix (`:56`) now use `SyntaxNodeUtilities.ContainsCommentOrDirective`, and `LineBreakListRewriter.CanSafelyCollapseBracketedArguments` — the formatter twin, which still held an inline copy of the narrow predicate — was redirected to Core with it. The diagnostic is no longer reported for the documented shape, so there is nothing left that the fix cannot clear.

RH5306 was checked for the same defect and had it; it is fixed here too. RH5109 was checked and does not: the formatter reshapes a documented parameter list without losing the comment, so blocking there would have withheld a fix that works.

---

Reihitsu.Core/SyntaxNodeUtilities.cs
thoenissen: Addressed, though not by the rename alone. Auditing every call site showed the choice is not binary — the predicate has to match both the trivia set and the *scope* of what each rewrite touches. `SyntaxNodeUtilities` now carries four predicates, each documented against the others:

- `ContainsCommentOrDirective` — full trivia; the default for a rewrite that can discard trivia.
- `InteriorContainsCommentOrDirective` — the node's own span. Needed because the full-trivia form also counts the member's documentation comment, which sits in the leading trivia of an attribute list; using it there made a documented class lose a fix an undocumented one still got.
- `GroupInteriorContainsCommentOrDirective` — a sibling group plus the gaps between them, the trailing trivia of the last sibling, and any directive touching either boundary. A merge that omitted the boundaries emitted `[In, Out] #endif`, which does not compile.
- `ContainsNonDocumentationCommentOrDirective` — kept, and named for what it omits. RH5109 and RH3005 need it: the formatter reshapes those shapes across a documentation comment without losing it, so blocking would withhold a working fix.

Core tests pin every scope explicitly, including a documented-vs-undocumented pair.

---

Reihitsu.Formatter/Pipeline/LineBreaks/PropertyLayoutLineBreakRewriter.cs
thoenissen: Verified fixed. `reihitsu-format` over the original counterexample now keeps the comment (`/// <summary>` / `/// note` / `/// </summary>` above `get;`) and the second pass reports no changes, so the guard is idempotent as well as non-destructive.

---

Reihitsu.Core/SyntaxNodeUtilities.cs
thoenissen: Verified fixed: `ContainsNonDocumentationCommentOrDirective` names the omission, `ContainsCommentOrDirective` is the unmarked default, and the two new scope-specific predicates are documented against each other. The IntelliSense trap is gone.

---

Reihitsu.Analyzer/Rules/Layout/RH5307IndexerBracketedArgumentsShouldBeSingleLinedAnalyzer.cs
thoenissen: RH5307 verified fixed: analyzer and code fix both use `ContainsCommentOrDirective`, and `LineBreakListRewriter.CanSafelyCollapseBracketedArguments` now calls the Core predicate instead of its inline copy.

The RH5109 conclusion in this reply is wrong and the finding stands — see the review summary. RH5109's fix never calls the formatter; `ApplyCodeFixAsync` rebuilds the list from `sourceText.ToString(parameter.Span)` and writes it back with `WithText`, so what the formatter does with a documented parameter list does not apply. Applying the fix to the exact source of the new `VerifyDocumentationCommentedParameterListIsOfferedACodeFix` test at `7ec710a` still deletes the comment.

---

Reihitsu.Analyzer.CodeFixes/Rules/Clarity/RH3005UseReadableConditionsCodeFixProvider.cs
```diff
@@ -174,7 +174,7 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
                 var binaryExpression = root.FindToken(diagnostic.Location.SourceSpan.Start).Parent?.AncestorsAndSelf().OfType<BinaryExpressionSyntax>().FirstOrDefault();
 
                 if (binaryExpression != null
-                    && SyntaxNodeUtilities.HasCommentsOrDirectives(binaryExpression) == false
+                    && SyntaxNodeUtilities.ContainsNonDocumentationCommentOrDirective(binaryExpression) == false
```
thoenissen: The swap does not carry the trivia along, so keeping the narrow predicate here loses a documentation comment. Verified at `7ec710a` by applying the fix to the source of the new `DocumentationCommentedYodaConditionIsOfferedACodeFix` test:

```csharp
return 0 < /** The counter. */ count;
```

becomes `return count > 0;` — `/** The counter. */` is deleted. The new test asserts only `Assert.IsNotEmpty(actions)`, so it pins that the action is offered, not that applying it preserves the comment, and the loss passes CI. Switch the guard to `SyntaxNodeUtilities.ContainsCommentOrDirective` and assert on the applied output.

---

Reihitsu.Analyzer.CodeFixes/Rules/Layout/RH5109ParametersMustBeOnSameLineOrSeparateLinesCodeFixProvider.cs
```diff
@@ -96,7 +96,7 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
             var parameterList = root.FindToken(diagnostic.Location.SourceSpan.Start).Parent?.FirstAncestorOrSelf<ParameterListSyntax>();
 
             if (parameterList == null
-                || SyntaxNodeUtilities.HasCommentsOrDirectives(parameterList))
+                || SyntaxNodeUtilities.ContainsNonDocumentationCommentOrDirective(parameterList))
```
thoenissen: Still open — the rename kept the narrow predicate and the fix still deletes documentation comments. The justification does not apply to this rule: `ApplyCodeFixAsync` never calls the formatter, it rebuilds the list from `sourceText.ToString(parameter.Span)` (`:51`, `:61`) and writes it back with `WithText` (`:68`), and leading trivia is outside every `parameter.Span`. Verified at `7ec710a` on the source of the new `VerifyDocumentationCommentedParameterListIsOfferedACodeFix` test:

```csharp
void Method(int first, int second,
            /// <summary>Third parameter.</summary>
            int third)
```

→ three parameters on separate lines, comment gone. That test asserts only `Assert.IsNotEmpty(actions)`, so it now locks the loss in. Use `SyntaxNodeUtilities.ContainsCommentOrDirective` and assert on the applied output. The stale claim at `:64-65` that the registration rejects lists carrying comments needs correcting with it.

---

thoenissen (commented):
Re-review of `7ec710a`. Seven of the nine prior findings are verified fixed and their threads are resolved — the four-predicate split with scope-specific names is a better answer than the rename I suggested, and the `GroupInteriorContainsCommentOrDirective` boundary-directive handling closes a case I had not raised.

Two of the three call sites deliberately kept on the narrow predicate do lose a documentation comment. Both are flagged inline with a verified counterexample: RH5109 (prior finding, still open) and RH3005 (new). In both the rationale is that the formatter reshapes the shape without losing the comment — but neither fix routes through the formatter, and both new tests assert only that the action is offered, never that the applied output preserves the comment, so the loss passes CI.

**Prior finding still open**

1. The Follow-up section still lists only RH5308 and RH5405 for the inline single-line comparison. `Reihitsu.Analyzer/Rules/Layout/RH5408SimpleAutoPropertiesShouldBeSingleLinedAnalyzer.cs:181-183`, `Reihitsu.Formatter/Pipeline/SwitchCaseBraces/SwitchCaseBraceRewriter.cs:198-203` (a named private copy of the kind issue #420 enumerates) and `Reihitsu.Analyzer/Rules/Layout/RH5406BracesMustNotBeOmittedFromMultiLineChildStatementsAnalyzer.cs:72-74` still carry it, unchanged across all three reviews. Convert RH5408:181-183 to `SyntaxNodeUtilities.IsSingleLineSpan(propertyDeclaration.SyntaxTree, relevantSpan)` and list the full set.

**Note on CI**

The last three commits are `[skip ci]`, so no check run has executed against `7ec710a`; the newest results on this PR are from `354cbd2`. The changed projects build and the targeted tests run locally, but the suite has not been verified at this head.

---

Reihitsu.Analyzer.CodeFixes/Rules/Clarity/RH3005UseReadableConditionsCodeFixProvider.cs
thoenissen: Addressed — the finding was right and my earlier justification was wrong. `BuildSwappedExpression` reparses both operands with `WithoutTrivia()`, so nothing in the comparison's interior survives; the formatter is never involved.

Reproduced first: applying the fix to `return 0 < /** The counter. */ count;` produced `return count > 0;` with the comment gone. The guard at `:177` now uses `SyntaxNodeUtilities.ContainsCommentOrDirective`, so the action is no longer registered for that shape.

The test that pinned the loss is replaced by two: `DocumentationCommentedYodaConditionIsNotOfferedACodeFix` asserts the registration refuses, and `DocumentedMemberWithYodaConditionIsFixedAndKeepsItsComment` applies the fix to a documented member and asserts the applied output still carries `/// <summary>Checks the counter.</summary>` — the comment sits in the member's leading trivia, outside the comparison, so that fix must keep working (2f369e7).

---

Reihitsu.Analyzer.CodeFixes/Rules/Layout/RH5109ParametersMustBeOnSameLineOrSeparateLinesCodeFixProvider.cs
thoenissen: Addressed. My previous reply generalised from the formatter to a fix that never calls it; you are right that `ApplyCodeFixAsync` rebuilds from `sourceText.ToString(parameter.Span)` and writes back with `WithText`, and that leading trivia lies outside every `parameter.Span`.

Reproduced first at `7ec710a`: the applied output was three parameters on separate lines with `/// <summary>Third parameter.</summary>` deleted. The guard at `:99` now uses `SyntaxNodeUtilities.ContainsCommentOrDirective`, and the stale claim at `:63-67` is corrected to state why the span-based rebuild forces that guard.

`VerifyDocumentationCommentedParameterListIsNotOfferedACodeFix` replaces the assertion that locked in the loss, and `VerifyDocumentedMemberStillGetsTheCodeFixAndKeepsItsComment` applies the fix to a documented member — the comment is leading trivia of the return type, outside the list — and asserts it survives (2f369e7).

---

Reihitsu.Analyzer.CodeFixes/Rules/Clarity/RH3005UseReadableConditionsCodeFixProvider.cs
thoenissen: Verified fixed at `ddda684`: the guard uses `ContainsCommentOrDirective`, and applying the fix to `return 0 < /** The counter. */ count;` now registers no action. The replacement tests assert the right properties — refusal for the interior comment, applied output for the documented member.

---

Reihitsu.Analyzer.CodeFixes/Rules/Layout/RH5109ParametersMustBeOnSameLineOrSeparateLinesCodeFixProvider.cs
thoenissen: Verified fixed at `ddda684`: the guard uses `ContainsCommentOrDirective`, applying the fix to the documented parameter list now registers no action, and the documented-member test confirms the working case still gets its fix. The `:63-67` comment is corrected.

---

Reihitsu.Analyzer.CodeFixes/Base/TargetAttributePlacementCodeFixProviderBase.cs
```diff
@@ -138,8 +138,17 @@ private bool TryGetFixableAttributeList(SyntaxNode root, Diagnostic diagnostic,
...
+        {
+            return false;
+        }
+
+        // Both placements rewrite the gap between the closing bracket and the member token: the fix overwrites the
+        // bracket's trailing trivia and clears the member token's leading trivia, so trivia in that gap is destroyed
+        // whichever placement is resolved. The interior guard above cannot see it, because it belongs to neither the
+        // attribute list's own span nor the member's. The formatter preserves such a comment, so the fix stays
+        // unregistered rather than reshaping the source destructively
+        if (SyntaxTriviaUtilities.WouldJoinAcrossUnjoinableTrivia(attributeList.CloseBracketToken, tokenAfter))
```
thoenissen, line 151: Extending this guard to both placement modes breaks parity with the analyzer and produces a diagnostic the user cannot clear. `TargetAttributePlacementAnalyzerBase.cs:78-79` still gates its suppression on `expectedMode == TargetAttributePlacementMode.SingleLine`, so in `SeparateLine` mode the diagnostic is reported while this fix now registers nothing.

Verified at `ddda684` on the source of the new `VerifyCommentBetweenAttributeListAndMemberIsNotOfferedACodeFix` test:

```csharp
[First] /* keep me */ internal class Example { }
```

RH5505 reports and no code action is registered. That test uses `GetCodeFixActionsAsync`, which synthesizes the diagnostic instead of running the analyzer, so it cannot see the break. Drop the `expectedMode == SingleLine` condition from `isBlockedSingleLineJoin` so the analyzer refuses on the same gap trivia, and add a `Verify` test asserting no diagnostic for this shape.

---

thoenissen (commented):
Re-review of `ddda684`. Both remaining findings from the last round are verified fixed and their threads are resolved: RH3005 and RH5109 refuse the documented shape, their replacement tests assert the applied output rather than mere registration, and the `ContainsNonDocumentationCommentOrDirective` doc now names the span-rebuild/reparse case that made both wrong. The RH5306 uncertainty from last round is closed by `VerifyCodeFixForListPatternWithDocumentationCommentPreservesTheComment`, and RH5408 no longer inlines the single-line comparison. CI is green on this head.

One new finding, flagged inline: commit `5d70e50` extended the attribute-placement gap guard to both placement modes in the code fix but not in the analyzer, so `SeparateLine` mode now reports a diagnostic that has no fix. It is a regression this PR introduces, not a pre-existing gap.

Everything else still open is pre-existing and already enumerated in the PR's own Follow-up section — the inline single-line comparisons (RH5308, RH5405, RH5406, `SwitchCaseBraceRewriter.SpansMultipleLines`), the `BlankLineEditor.cs:204` region-directive match, the `TargetAttributeListShapeAnalyzerBase.cs:77` parity gap, `FieldDeclarationSplitTransform.cs:83`, and the two documentation-comment formatting defects. None of them are regressions from this PR and none need to block it.

The RH5111 deviation from issue #420's "keep behaviour unchanged" requirement still needs the explicit sign-off the PR body asks for.

---

Reihitsu.Analyzer.CodeFixes/Base/TargetAttributePlacementCodeFixProviderBase.cs
thoenissen: Fixed at `18a9177`. The finding was right: fourteen of the fifteen rules deriving from `TargetAttributePlacementAnalyzerBase` resolve to `SeparateLine` (only RH5523 defaults to `SingleLine`, and RH5530 resolves to `SingleLine` only for a single-line property accessor), so the analyzer kept reporting where this fix registers nothing.

`TargetAttributePlacementAnalyzerBase.cs:78-79` no longer gates the suppression on the placement mode. The local is renamed `isBlockedByGapTrivia` and now holds `WouldJoinAcrossUnjoinableTrivia(attributeList.CloseBracketToken, tokenAfter)` alone, with the comment stating that both placements rewrite that gap — the analyzer refuses on exactly the condition this guard refuses on.

Reproduced first, per `CLAUDE.md`. Two analyzer-side `Verify` tests were added and both failed against `ddda684` with `Mismatch between number of diagnostics returned, expected "0" actual "1"`:

- `VerifyNoDiagnosticWhenCommentSitsBetweenAttributeListAndMember` (RH5505) on `[First] /* keep me */ internal class Example { }`.
- `VerifyNoDiagnosticWhenCommentSitsBetweenAttributeListAndAccessor` (RH5530) on `[First] /* keep me */ get;` inside a multi-line property, which pins the same shape reached through the virtual `ResolvePlacementMode` override rather than the default.

`VerifyCommentBetweenAttributeListAndMemberIsNotOfferedACodeFix` is kept as-is; as you noted, its `GetCodeFixActionsAsync` path synthesizes the diagnostic and cannot see the analyzer side, so the two tests cover different halves.

The existing interior-comment behaviour is untouched: `[First /* keep */] internal class Example { }` puts the comment in the identifier's trailing trivia, not in the bracket-to-member gap, so `VerifyDiagnosticWithoutCodeFixWhenCommentsArePresent` still reports the diagnostic without offering the fix. Full `Reihitsu.Analyzer.Test` run is green (2241 passed).

Leaving this unresolved for you.

---

thoenissen: Verified fixed at `bde76e0`. `isBlockedByGapTrivia` no longer gates on the placement mode, and the probe that previously failed with "No code fix action was registered for the diagnostic" now fails with "The analyzer did not report a fixable diagnostic" — the analyzer and the fix refuse on the same condition. All 252 tests across the fifteen placement rules and every rule this PR touches pass locally, as do the Core and formatter documentation-comment suites after the `main` merge.

Worth recording: the fix's `WouldJoinAcrossUnjoinableTrivia` check is unconditional on `main` too (`TargetAttributePlacementCodeFixProviderBase.cs:142`), so this closes a parity gap that predates the PR rather than one it introduced.

The interior-comment case stays as it is and is out of scope here: `[First /* keep */] internal class Example { }` still reports without offering a fix, because the analyzer has no interior guard to match `InteriorContainsCommentOrDirective`. That is the placement-base twin of the `TargetAttributeListShapeAnalyzerBase.cs:77` gap already listed under Follow-up, is present on `main` with `VerifyDiagnosticWithoutCodeFixWhenCommentsArePresent` pinning it, and is not widened here.

---

PR #580, opened 2026-07-28
Add opt-in UTF-8 BOM normalization to the CLI

thoenissen:
## Summary

Adds opt-in `--utf8-bom` normalization to the CLI, including encoding-only change detection for normal, check, and dry-run modes. Supported Unicode inputs are rewritten as UTF-8 with a byte order mark while the default path continues to preserve detected encodings.

## Why

Repositories can enforce one on-disk source encoding during their existing formatting pass without a separate conversion tool or a behavior change for users who do not enable the option.

## Linked issues

Closes #477

## Review notes

Encoding-only dry runs report the affected file without a textual diff because the decoded source is unchanged. Runs above the confirmation threshold stream only enough decoded text to classify each candidate through the formatter's shared generated-source policy, so skipped generated files do not trigger a large-run prompt without retaining repository-wide syntax trees. Accepted runs then read current file content normally, preserving edits made while confirmation is pending.

## Follow-up work

None.


Reihitsu.Cli/FormatCommandHandler.cs
```diff
@@ -323,30 +361,79 @@ private async Task HandleChangedFileAsync(string filePath, string originalConten
...
+    /// Determines whether a file is marked as auto-generated for large-run planning
+    /// </summary>
+    /// <param name="filePath">The file path to inspect</param>
+    /// <param name="cancellationToken">A token to monitor for cancellation requests</param>
+    /// <returns>A task containing <see langword="true"/> when the file is marked as auto-generated; otherwise, <see langword="false"/></returns>
+    private async Task<bool> IsAutoGeneratedFileAsync(string filePath, CancellationToken cancellationToken)
+    {
+        try
+        {
+            var fileRead = await _fileSystem.ReadFileAsync(filePath, cancellationToken).ConfigureAwait(false);
```
thoenissen: Every unforced normal run above the 25-file threshold now reads and parses each candidate here, then `ProcessFileAsync` repeats both operations after confirmation, roughly doubling I/O and parsing even when `--utf8-bom` is disabled. Cache the pre-scan result for processing or inspect only the generated-source header.

thoenissen: Addressed: large-run planning now streams only enough decoded text to classify the generated-source header, then normal processing performs one fresh full read/parse after confirmation. This avoids duplicate full-file I/O and parsing without retaining stale syntax trees; prompt-boundary edits, all supported encodings, and split comment delimiters are covered by regressions (37c993a7, 841bc2f0).

thoenissen: Verified: large-run planning now streams only the source prefix needed for generated-file classification, while accepted files perform one fresh full read and parse.

PR #586, opened 2026-07-29
Format complex collection elements recursively

thoenissen:
## Summary
Formats multi-line complex collection elements with their own brace level and recursively aligned key/value expressions while preserving short single-line pairs. Extends RH5303 and RH5304 with shared recursive layout analysis so analyzer diagnostics, code fixes, and formatter output agree for ordinary and assignment-owned nested collection initializers.

## Why
Nested object initializers can become multi-line during formatting, but their containing dictionary-style element previously lacked matching line-break and indentation handling. The analyzer rules also need to reject every layout the formatter would change without reporting formatter-stable trivia layouts.

## Linked issues
Closes #474

## Review notes
Complex elements expand only after their contents span multiple lines. The enclosing collection braces remain anchored to the `new` keyword, short `{ key, value }` entries remain inline, and comments, directives, and disabled text retain their positions. RH5303 and RH5304 recursively validate brace and expression anchors, line ownership, standalone and same-line comments, and documentation-comment continuation markers. Shared documentation-comment alignment keeps full-document formatting and scoped code fixes convergent.

## Follow-up work
None.


Reihitsu.Analyzer/Rules/Layout/RH5304NestedCollectionInitializerAssignmentsShouldBeFormattedCorrectlyAnalyzer.cs
```diff
@@ -135,6 +135,20 @@ private void CheckNestedCollectionInitializerAssignment(SyntaxNodeAnalysisContex
             || expressionLinePositions.Select(position => position.Line).Distinct().Count() != expressionLinePositions.Length)
         {
             context.ReportDiagnostic(CreateDiagnostic(assignment.GetLocation()));
+
+            return;
+        }
+
+        foreach (var complexElement in collectionInitializer.Expressions
```
thoenissen: RH5304 still accepts misindented ordinary elements because this new column validation visits only complex-element nodes. With aligned braces and one value per line, `Values = { ... }` can contain values at arbitrary columns without a diagnostic even though the formatter moves them to `openBrace + 4`; validate every outer expression anchor before recursing into complex elements.

thoenissen (commented):
1. **[medium] Initializer comment parity remains incomplete.** `RH5303.CheckBraces`, `RH5304`'s outer-brace comparison, and counterpart `RH5301` still compare raw token columns, so formatter-stable forms such as `/* keep */ {`, `/* keep */ }`, and `/* keep */ A = 1` report diagnostics; reuse the new comment-aware anchor check for these outer braces and object members.

Reihitsu.Analyzer/Rules/Layout/RH5304NestedCollectionInitializerAssignmentsShouldBeFormattedCorrectlyAnalyzer.cs
thoenissen: Addressed: RH5304 now validates every multiline collection element anchor before recursively checking complex elements, with code-fix, Fix All, and formatter-parity regressions (4f9edf55).

thoenissen: Verified: RH5304 now checks every multiline collection element anchor before recursive complex-element validation, and the added fix, Fix All, and parity regressions cover the defect.

Reihitsu.Analyzer/Core/InitializerLayoutAnalysisUtilities.cs
```diff
@@ -0,0 +1,130 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+using Reihitsu.Core;
+
+namespace Reihitsu.Analyzer.Rules.Layout;
+
+/// <summary>
+/// Shared layout analysis for initializer anchors and multi-line complex elements
+/// </summary>
+internal static class InitializerLayoutAnalysisUtilities
```
thoenissen, line 14: [medium] This non-`RHdddd` helper resides in `Reihitsu.Analyzer.Rules.Layout`, so `RulesTests.OtherTypesShouldNotBeInRulesNamespace` fails on the current head, as the SonarCloud job demonstrates. Move the helper to a non-`Rules` namespace and directory, then update its consumers so the architecture suite passes.

PR #589, opened 2026-07-30
Add gh-rubber-duck Behavior Contract gate to the GitHub workflows

thoenissen:
## Summary

Adds a reusable, read-only `gh-rubber-duck` skill and `/gh-rubber-duck` command that turn an issue (or a PR under behavioral review) into an explicit Behavior Contract before code is written, and reworks the orchestrator workflows around it. `gh-implement` now triages each run as routine or behavioral, and on a behavioral run reads the skill in the parent and runs exactly one fresh read-only Rubber Duck subagent before any production or test edit, then derives its regression matrix from the accepted contract. `gh-implement` and `gh-apply-review` gain a local self-review, move `origin/main` synchronization ahead of the official preflight, cap the official preflight at one attempt plus one retry around a single consolidated repair cycle, and run the full validation once. `gh-preflight` is documented as the final gate with explicit calling conditions; its analytical coverage is unchanged. All four workflows now name the repository's existing test helpers instead of describing the coverage abstractly. Everything is mirrored for Codex (`.codex/`) and Claude (`.claude/`), including the Codex `agents/openai.yaml` metadata, and documented in `AGENTS.md` and `CLAUDE.md`.

## Why

Repeated preflight rounds, late discovery of behavioral requirements, and full-suite reruns dominated the cost of recent issue runs. Most of that rework came from implementing against an unstated contract and from discovering problems one preflight at a time, so the redesign pins the contract down first, makes the cheap local check carry the discovery load, and gives the expensive official gate a hard budget on a head that is already synchronized with `main`.

The scope triage keeps that from taxing small work: a run that changes no analyzer, formatter, code-fix, or `Reihitsu.Core` behavior, admits one reading, stays in a small listable file set, and is not a bug report may replace the Rubber Duck subagent with a short contract note, and may skip the official preflight when the diff contains no production code. Trip-wires void the classification, and the run report always names it.

The helper map exists because the coverage the skills were asking for is already implemented in this repository — repo-wide formatter idempotency, LF/CRLF assertion helpers, parity bases, code-fix convergence through the Roslyn verifier — but the skills never named the helpers, so the stronger ones stayed largely unused.

## Linked issues

_None._ Follow-up filed as #596 (enforce that coverage through meta-tests rather than instructions).

## Review notes

- Workflow/documentation only — no analyzer, formatter, CLI, or other product code is touched.
- Behavior preserved deliberately: issue ownership and draft-PR claiming, branch naming, `[skip ci]` intermediate commits with a single final CI trigger, replying to review threads without resolving them, exact-head and clean-worktree checks, test-first bug fixes, formatter-before-tests, and fresh independent preflight agents.
- Stale text permitting unlimited fix/preflight loops was removed rather than contradicted; a grep for that phrasing and for post-preflight `main` merges now comes back empty.
- Two compatibility decisions worth a look: the `gh-implement` frontmatter switched to the folded `>-` style already used by `gh-apply-review`, because strict YAML truncated the description at `"implement #123"`; and the four test projects now run with `--no-build` after the Release build that immediately precedes them, with instructions to drop the flag if anything changed since.
- Validation was static, as the change is instructional: referenced paths resolve, section order is local self-review → base sync → official preflight → full validation in all four orchestrator skills, all skill frontmatter parses, and the acceptance scenarios were walked against the written instructions.

## Follow-up work

- #596 — meta-tests that enforce idempotency and Fix All coverage, so the helpers named here are also required by the suite.


thoenissen (commented):
Confirmed workflow findings are inline.

PR #597, opened 2026-07-31
Make brace placement and expression-body conversion converge in one formatting pass

thoenissen:
## Summary

Place the accessor-list opening brace of indexer and event declarations during the first formatting pass instead of the second, and close the rest of the same defect class found along the way: `unsafe` and labeled statement blocks, region pairs straddling an element-body boundary, and expression bodies carrying a directive the rewrite cannot re-host.

Three decisions now have a single owner: `BracePlacer.NormalizeOwnedBraces` for brace placement, `LineBreakDetection.ShouldNormalizeAccessorListBraces` for whether an accessor list gets Allman braces, and `ExpressionBodyRewriteUtilities.BlocksRewrite` for whether an expression body may be rebuilt as a block.

## Why

Closes #581.

One mechanism explains the reported bug and most of what this branch fixes: **a brace is normalized on a node that does not contain the token preceding it**. A same-pass descendant edit detaches that node, the anchor cannot be resolved, and the brace silently stays where it was until a second pass moves it. Properties, methods and constructors escaped this because they already get the same normalization at declaration level; indexers, events, `unsafe` blocks and labeled blocks did not.

A sweep of every statement kind carrying a contained block now shows all thirteen converging in a single pass.

## Linked issues

Closes #581

## Review notes

- **The issue's root-cause attribution is incorrect.** It points at `ExpressionBodiedIndexerTransform`, but a hand-written indexer that never had an expression body reproduces the identical wrong first pass. `AccessorListDeclarationBraceTests` pins that case, so the tests target the mechanism rather than the reported symptom. The transforms *are* changed here, but for a separate defect (below).
- **Event declarations are fixed alongside indexers**, as one defect class. This changes formatter output for a construct the issue does not mention.
- **The accessor-list guard is interior-scoped.** A comment or directive inside the list forces normalization, mirroring `CanCollapseAutoPropertyToSingleLine`; a comment trailing the closing brace sits outside the list and must not split it. A multi-line auto-accessor indexer still keeps its brace on the declaration line where the property counterpart collapses — pre-existing and deliberately unchanged.
- **Straddling region pairs are removed as a unit.** Previously only the half inside the element body was deleted, orphaning its partner and turning compiling source into `CS1028`. Removing both matches `RH7303DoNotPlaceRegionsWithinElementsCodeFixProvider`, so the formatter and that code fix now agree.
- **Expression-body conversion is refused for three directive placements**: any directive between the arrow and the expression (it would be re-hosted directly after the generated `return`, where a directive cannot appear); a conditional group split by the rewrite; and any region directive reaching the rewrite. The region case is the widest of the three on purpose — see below. A `#pragma` or a balanced conditional after the expression still converts, so the pre-existing issue #422 behavior is unchanged.
- **RH3202 and RH3203 share that predicate**, so neither reports a member whose code fix the formatter would refuse. Both rule docs gained an `Exceptions` section. This matters because `RH3202` is `Action="Error"` in `Release.ruleset`, so a reported-but-unfixable member would have been an unclearable build error.

## Follow-up work

Both captured as local drafts under `plans/issues`, for separate pull requests:

- **Generated `return` has no separator and can re-lex as an identifier.** `ExpressionBodyToBlockConverter` emits `return` without separating trivia and relies on `HorizontalSpacingPhase` to repair it, but the region phase reparses changed text before that phase runs, so `return 1` re-lexed as `return1` — destroying the member, dropping directives, corrupting sibling members in the same document, and throwing on the code-fix path. Pre-existing on `main`. This PR's region refusal keeps the formatter out of that state; it does not remove the fragility, and once the separator is fixed the region refusal can be narrowed again.
- **Sibling accessors sharing a line** (`} remove`, `} set`) are never split. Pre-existing, and new formatter behavior rather than a bug fix. `} catch`, `} finally` and `} else` are the statement-level members of the same class.

Not yet captured: an expression body with a leading or trailing line comment still needs two passes to settle — pre-existing, same "output depends on run count" family as #581, different cause.


Reihitsu.Formatter/Pipeline/LineBreaks/DeclarationBraceLineBreakRewriter.cs
```diff
@@ -90,6 +91,33 @@ private static TNode CollapseParameterListToDeclarationLine<TNode>(TNode node,
...
+    /// test, which would otherwise report an empty list as an auto-accessor list.
+    /// </remarks>
+    private static bool ShouldNormalizeAccessorList(AccessorListSyntax accessorList)
+    {
+        if (accessorList == null || accessorList.OpenBraceToken.IsMissing)
+        {
+            return false;
+        }
+
+        if (accessorList.Accessors.Count == 0)
```
thoenissen: The guard keys on `Accessors.Count == 0`; the property counterpart keys on `SyntaxNodeUtilities.ContainsCommentOrDirective` (`PropertyLayoutLineBreakRewriter.CanCollapseAutoPropertyToSingleLine`). The count check both under- and over-triggers.

**Under-triggers.** An auto-accessor list that keeps at least one parsed accessor next to a comment or directive stays `IsAutoPropertyAccessorList == true`, so the brace is never normalized and stays on the declaration line permanently — exactly the state the remark above says this guard exists to prevent. Verified on this branch:

```csharp
interface I
{
    int Value {
        // note
        get;
        set;
    }

    int this[int index] {
        // note
        get;
        set;
    }
}
```

produces

```csharp
interface I
{
    int Value
    {
        // note
        get;
        set;
    }

    int this[int index] {
        // note
        get;
        set;
    }
}
```

The same divergence reproduces with `#if DEBUG` around one accessor instead of the comment.

**Over-triggers.** A genuinely empty `{ }` carries no comment or directive, so the property keeps it on the declaration line while the indexer and event now move it — a divergence this PR introduces and pins with `IndexerBraceMovesWhenAccessorListIsEmpty`.

Replacing the count check with `SyntaxNodeUtilities.ContainsCommentOrDirective(accessorList)` closes both and keeps `IndexerBraceMovesWhenAccessorsAreInInactiveBranch` passing.

---

Reihitsu.Formatter/Pipeline/LineBreaks/DeclarationBraceLineBreakRewriter.cs
```diff
@@ -130,20 +158,29 @@ private BaseTypeDeclarationSyntax NormalizeTypeBraces(BaseTypeDeclarationSyntax
...
+    /// <param name="getCloseBrace">Selects the close brace token from the current node</param>
+    /// <returns>The node with normalized braces</returns>
+    /// <remarks>
+    /// The normalization runs on the owning declaration rather than on the braced node itself,
+    /// because the token preceding the open brace lives outside that node. A rewriter that visits
+    /// the braced node alone receives a detached node whenever one of its descendants changed
+    /// during the same pass, so the anchor cannot be resolved and the brace stays put. Each token is
+    /// re-selected from the current node between steps, because every step returns a new node.
+    /// </remarks>
+    private TNode NormalizeOwnedBraces<TNode>(TNode node,
```
thoenissen: `NormalizeOwnedBraces` is now the third copy of the same three-step brace policy: `PropertyLayoutLineBreakRewriter.VisitPropertyDeclaration` repeats it verbatim in both branches of its `if`/`else` (lines 301-303 and 308-310 are byte-identical), and `BracePlacer.NormalizeContainedBlock` repeats it annotation-based. This PR generalized the helper to arbitrary token selectors but left the property copies in place, so accessor-list brace policy has three owners that can drift.

Route the property accessor list through this helper and collapse the identical `if`/`else` branches.

---

Reihitsu.Formatter.Test/Regression/LineBreaks/AccessorListDeclarationBraceTests.cs
```diff
@@ -0,0 +1,306 @@
...
+                                     set;
+                                 }
+                             }
+                             """;
+
+        // Act & Assert
+        AssertRuleResult(input);
+    }
+
+    #endregion // Methods
```
thoenissen, line 436: No test places a comment in the gap this change edits (`]` or the event identifier to `{`) on the hand-written path; `ArrowOnOwnLineDoesNotDoubleTheLineBreak` reaches that gap only through the expression-body transform, which rebuilds the trivia instead of joining it. Add the direct case:

```csharp
public int this[int index] /* c */ {
    get { return 0; }
}
```

Current output, verified on this branch, is correct and should be pinned:

```csharp
public int this[int index] /* c */
{
    get
    {
        return 0;
    }
}
```

---

thoenissen (commented):
The declaration-level normalization is the right anchor and the LF/CRLF plus second-pass assertions in `AssertRuleResult` cover the issue's idempotence requirement. Three inline findings, plus two defects that have no line anchor in this diff.

Both of the following are listed as out-of-scope follow-ups in the PR description. Their severity is the reason they should not ship as follow-ups: one is the same defect class this PR closes, the other emits non-compiling output. There is no issue tracking either.

**1. `Reihitsu.Formatter/Pipeline/LineBreaks/LineBreakContainedBlockRewriter.cs` — `UnsafeStatementSyntax` missing from `TryGetContainedBlock` (same defect class as #581).**

`TryGetContainedBlock` lists `LockStatementSyntax`, `FixedStatementSyntax`, `CheckedStatementSyntax`, `TryStatementSyntax`, `CatchClauseSyntax` and `FinallyClauseSyntax`, but not `UnsafeStatementSyntax`. The mechanism is identical to the indexer case: the anchor token lives outside the block, so a block detached by a same-pass descendant edit loses the normalization and needs a second run. Verified on this branch — `unsafe` needs two passes while `lock` converges in one:

```csharp
unsafe void M()
{
    unsafe {
        if (true) { int x = 1; }
    }

    lock (this) {
        if (true) { int y = 1; }
    }
}
```

first pass:

```csharp
unsafe void M()
{
    unsafe {
        if (true)
        {
            int x = 1;
        }
    }

    lock (this)
    {
        if (true)
        {
            int y = 1;
        }
    }
}
```

Closing the class is one switch arm: `UnsafeStatementSyntax unsafeStatement => unsafeStatement.Block,`.

**2. Region phase deletes a `#region` opened between a member header and its brace — output does not compile (CS1028).**

The opening directive is dropped while both `#endregion` directives survive, and the deleted region's name is transferred onto the wrong `#endregion`. Verified on this branch and on `main`: the input parses with zero diagnostics, the output reports `CS1028 Unexpected preprocessor directive`.

```csharp
class C
{
#region Members

    public int this[int index]
#region Inner
    {
        get { return 0; }
    }
#endregion

#endregion
}
```

output:

```csharp
class C
{
    #region Members

    public int this[int index]
    {
        get
        {
            return 0;
        }
    }

    #endregion // Inner

#endregion // Members
}
```

Directive loss that turns compiling input into non-compiling output is a correctness defect, not a layout follow-up.

---

Reihitsu.Formatter/Pipeline/LineBreaks/DeclarationBraceLineBreakRewriter.cs
thoenissen: Fixed. `LineBreakDetection.ShouldNormalizeAccessorListBraces` now keys on `SyntaxNodeUtilities.InteriorContainsCommentOrDirective` and is shared by all three rewriters. Verified on `badd4db`: the commented auto-accessor indexer and its property counterpart now produce identical layout, the `#if DEBUG` variant matches, an empty `{ }` stays on the declaration line for both, and a comment trailing the closing brace no longer splits the list.

---

Reihitsu.Formatter/Pipeline/LineBreaks/DeclarationBraceLineBreakRewriter.cs
thoenissen: Fixed. The three-step policy now lives in `BracePlacer.NormalizeOwnedBraces`, and both `DeclarationBraceLineBreakRewriter` and `PropertyLayoutLineBreakRewriter` call it; the identical `if`/`else` branches are gone. `NormalizeContainedBlock` keeps its own annotation-based sequence, which is required because the block is a re-resolved child node.

---

Reihitsu.Formatter.Test/Regression/LineBreaks/AccessorListDeclarationBraceTests.cs
thoenissen: Fixed by `CommentBeforeAccessorListOpenBraceIsPreserved`, which pins the exact shape and output.

---

thoenissen (commented):
All five findings from the previous review are resolved. Verified on `badd4db`:

- `UnsafeStatementSyntax` is in `TryGetContainedBlock`; `unsafe { … }` with a reflowed descendant now converges in one pass and is stable on the second.
- The straddling region pair is removed as a unit. The original repro now parses with zero diagnostics and is stable, and the legitimate type-level `#region Members` pair is preserved. `RegionAroundMembersRemainsUnchanged` pins the negative case, and the behavior matches what `RH7303DoNotPlaceRegionsWithinElementsCodeFixProvider` does with the same input.
- `ShouldNormalizeAccessorListBraces` is a single predicate keyed on the list interior, shared by all three rewriters; the property/indexer divergences on comments, directives, and empty lists are gone.
- The brace policy has one owner in `BracePlacer.NormalizeOwnedBraces`.
- The comment-in-gap case is pinned.

I also traced the remaining members of the defect class — bare blocks, labeled/case blocks, `switch` — and all converge in one pass.

Two new findings, both without a changed-line anchor.

**1. high — An expression body split by a conditional directive is rewritten into non-compiling source and never stabilizes.**

This is the first item under "Follow-up work". Confirmed byte-identical on this branch and on the PR base `3516bbe`, so it is not introduced here — but it is a directive relocation, not a layout issue, and it reaches the same `ExpressionBodyToBlockConverter` path that this PR's `ExpressionBodiedIndexerTests` exercise.

```csharp
class C
{
    public int M() =>
#if DEBUG
        1;
#else
        0;
#endif
}
```

The input parses with zero diagnostics. First-pass output:

```csharp
class C
{
    public int M()
    {
        return #if DEBUG
        1;
#else
        0;
    }
#endif
}
```

`#if DEBUG` is moved into the middle of the `return` statement and `#endif` is pushed outside the member, producing `CS1040` plus two `CS1028`. The second pass produces different output again, so the result also depends on how many times the formatter runs — the same property issue #581 was filed about. The identical shape on an indexer behaves the same way.

The minimum safe behavior is to refuse the transform when the expression body's span carries a conditional directive, the way other guards in this pipeline refuse an unjoinable edit.

**2. low — Sibling accessors that share a line are never split.**

The second "Follow-up work" item; also confirmed identical on `3516bbe`.

```csharp
public event System.EventHandler Changed {
    add { _handler += value; } remove { _handler -= value; }
}
```

produces

```csharp
public event System.EventHandler Changed
{
    add
    {
        _handler += value;
    } remove
    {
        _handler -= value;
    }
}
```

The output compiles and is stable; `remove` keeps sharing a line with the preceding closing brace. `LineBreakBlockRewriter.EnsureStatementsStartOnSeparateLines` has no accessor-list counterpart.

---

Issue #699, opened 2026-08-20
[FORMATTER QUESTION] A comment above the first invoked link pins a chain at block indentation rather than under its root


thoenissen:
### Description

When a comment sits directly above a method chain's first invoked link and the chain also wraps a non-invoked prefix dot, the resulting layout places every link at **block indentation** rather than under the chain root:

```csharp
        var x = a
        .Prop

        // keep wrapped
        .Foo()
        .Bar();
```

The chain root `a` is at column 16, while `.Prop`, `.Foo()` and `.Bar()` sit at column 8.

This is pre-existing behavior — verified byte-identical before and after PR #691 — and it is `RH5201`-clean, so it is not a rule violation.

It is raised because PR #691 added a regression test, `MethodChainAlignmentTests.CommentAboveFirstInvokedLinkKeepsWrappedPrefixDotUncollapsed`, to pin the collapse-suppression boundary so that the narrowed comment bail could not drift. That test turns previously undocumented behavior into an asserted baseline, and the layout it asserts resembles the complaint in issue #685 — "the member name stays on its own continuation line, indented to block level rather than joined to the preceding dot or aligned to any chain anchor".

### Question to settle

Is block indentation the intended end state for this shape, or should such a chain align under its root the way a comment-exempt chain without a wrapped prefix dot does?

- If block indentation is intended, this can be closed with a note on the existing test recording that the layout is deliberate.
- If it is not, the alignment phase's treatment of this shape needs a decision, and the pinned test's expectation changes with it.

### Logs or Additional Context

- Raised as a `gh-preflight` hint during PR #691; confirmed pre-existing by comparing against `origin/main`.
- Related to the sibling issue about which column comment-exempt chains receive, but distinct: this one is RH5201-clean and is a design question rather than a rule violation.

PR #721, opened 2026-08-24
Collapse a wrapped null-forgiving prefix the same way as a plain prefix dot

### LINKED ISSUE #719: [FORMATTER BUG] A comment above the first invoked link pins a chain rooted in a wrapped null-forgiving prefix at block indentation
### Description

Issue #699 fixed the case where a comment sits above a chain's first *invoked* link while an earlier, uncommented, wrapped non-invoked prefix dot (e.g. `.Prop`) is the actual collapse candidate: that prefix now correctly collapses onto the chain-root line, and later links align to the resulting column.

A sibling shape was found and verified during that fix's preflight audit but deliberately left out of scope: the same block-indentation defect still reproduces when the wrapped prefix is introduced by a postfix null-forgiving operator (`!`) rather than a plain member-access dot. `ChainWalker.CollectAlignmentDots`/`CollectInvokedLinkDots` never collect a separate dot token for a directly-invoked member access on a null-forgiving receiver (`a!.Foo()`) — the `!` token stands in for the whole `!.Foo()` link (`FluentChainUtilities.GetInvokedLinkOperator`). When `!` is immediately followed by a non-invoked member access instead of an invocation (`a` ⏎ `!.Prop` ⏎ comment ⏎ `.Foo()` ⏎ `.Bar()`), the collapse/alignment machinery does not recognize `!.Prop` as a collapsible prefix the way a plain `.Prop` is recognized, and the whole chain stays pinned at block indentation under a root that may sit far to the right — the same symptom #699 fixed for the plain-dot case.

This is confirmed pre-existing (present on `main` before #699/#698's fix, `origin/main` @ `f5f357dabd884f72092a60617416b55eb04e10b1`) and confirmed still present after PR #718 merged the #699/#698 fix — both the original Behavior Contract's defect-class sweep and the official preflight audit that shipped PR #718 verified the output is byte-identical before and after that PR for this specific shape. It is `RH5201`-clean (the analyzer only compares invoked links' columns, and none differ here), so it is not a rule violation — the same character as #699 itself.

### Expected Behavior

Not determined by this report. Per the pattern #699 established, a plausible target is: the wrapped `!.Prop` prefix collapses onto the chain-root line (`a!.Prop`), and the following invoked links align to the resulting column — mirroring #699's resolution. But this needs its own Behavior Contract pass rather than assuming the #699 target applies unchanged, since `!.Prop`'s dot bookkeeping differs from a plain `.Prop`'s (see Logs below).

### Actual Behavior

The chain stays wrapped at block indentation, with every continuation line (`!.Prop`, the comment, and every invoked link after it) aligned to the enclosing block's indentation rather than under the chain root or any invoked link's column.

### Input Code

```csharp
class C
{
    void M()
    {
        var x = a
            !.Prop
            // keep wrapped
            .Foo()
            .Bar();
    }
}
```

### Expected Formatted Output

Not determined by measurement — see "Expected Behavior" above. Confirm the target via a Behavior Contract pass before implementing, the same way #699 was settled.

### Actual Formatted Output

```csharp
class C
{
    void M()
    {
        var x = a
        !.Prop

        // keep wrapped
        .Foo()
        .Bar();
    }
}
```

`a` sits at column 16 (inside a nested method body); `!.Prop`, the comment, `.Foo()` and `.Bar()` all sit at column 8. Stable under a second formatter pass (does not oscillate) and `RH5201`-clean under both LF and CRLF.

### CLI Version

1.0.0-rc2

### Operating System

Linux

### Logs or Additional Context

- Originally found as candidate "C20" during the Behavior Contract analysis for issues #699/#698 (see PR #718), and re-verified byte-identical on `main` both before and after that PR by the official preflight audit that shipped it.
- Related to #699, but distinct: `ChainLineBreakRewriter.FindFirstWrappedChainOperator`'s collapse-candidate search (the mechanism #699/#718 touched) is bounded to never return a dot positioned after the chain's first invoked link. For `a!.Foo()` (no intervening non-invoked member access), that bound correctly falls back to `!` itself, and #718 fixed several related `!`-rooted oscillation bugs as a side effect. For `a!.Prop.Foo()` specifically — where a non-invoked member access follows the null-forgiving operator before the first invocation — the collapse/alignment machinery does not treat `!.Prop` as a collapsible unit the way a plain `.Prop` is, which is why this shape needs its own analysis rather than a one-line extension of #718's fix.
- Derived from an adversarial fixture during analysis, not from a real-world code sample. Confirm the reproduction before implementing.

thoenissen:
## Summary

`FindFirstWrappedChainOperator`'s collapse-candidate search now picks the first chain-spine token that is itself wrapped onto its own line — including a postfix null-forgiving operator (`!`) — instead of unconditionally excluding every `!` token. `ChainWalker.DotHasIntermediateMemberAccess` now also checks a null-forgiving operator's own operand, the same way it already checked a plain dot's receiver.

- `Reihitsu.Formatter/Pipeline/LineBreaks/Rewriter/ChainLineBreakRewriter.cs`: generalized the collapse-candidate search from "first non-postfix-unary spine token" to "first spine token with leading end-of-line trivia."
- `Reihitsu.Formatter/Pipeline/LineBreaks/Utilities/ChainWalker.cs`: extended `DotHasIntermediateMemberAccess` to check a null-forgiving operator's own operand.
- `Reihitsu.Formatter.Test/Regression/Indentation/MethodChainAlignmentTests.cs`: added regression tests for the issue's reported shape, the no-comment collapse, the two-wrapped-prefix variant, and the intermediate-member-access guard-delta.

## Why

A wrapped `!` introducing a non-invoked member access (`a` ⏎ `!.Prop.Foo()`) was never offered to the chain collapse the way a plain wrapped prefix dot is (#699/#718's fix): the search excluded every postfix null-forgiving token outright, so the line break before `!` was never recognized, and the chain stayed pinned at block indentation whenever a comment sat above the first invoked link. #699/#718 deliberately scoped this sibling shape out; this closes it the same way.

## Linked issues

Closes #719

## Review notes

The search change is a generalization, not a special case for `!`: it reduces to the same outcomes as the old exclusion-based logic for every previously-tested shape (verified by tracing several representative inputs through `scripts/trace.sh` and by the full existing `MethodChainAlignmentTests` suite staying green), while now also picking up the `!`-prefix case.

Widening the search surfaced a second, related gap: `a.Prop1` ⏎ `!.Prop2.Foo()` is a fluent chain with an intermediate member access through `!`'s own operand, which `DotHasIntermediateMemberAccess` did not check before. Without the accompanying fix, this shape would have collapsed across the link where its plain-dot counterpart (`a.Prop1` ⏎ `.Prop2.Foo()`) correctly stays wrapped. The `NullForgivingPrefixWithIntermediateMemberAccessStaysWrapped` test pins this guard-delta.

Manually traced the issue's exact input through `scripts/trace.sh` and confirmed a stable (non-oscillating) second pass. Full suite (4,890 tests across all projects) passes.

## Follow-up work

None.

### DIFF
--- a/Reihitsu.Formatter/Pipeline/LineBreaks/Utilities/ChainWalker.cs
+++ b/Reihitsu.Formatter/Pipeline/LineBreaks/Utilities/ChainWalker.cs
@@ -257,9 +257,11 @@ internal static class ChainWalker
 
     /// <summary>
     /// Determines whether a single chain dot token has an intermediate member access between
-    /// the dot and the chain root
+    /// the dot and the chain root. A postfix null-forgiving operator is checked the same way, through
+    /// its own operand, since <c>a.Prop!.Foo()</c> belongs to the same fluent chain the plain-dot check
+    /// keeps wrapped for <c>a.Prop.Foo()</c> (issue #719)
     /// </summary>
-    /// <param name="dotToken">The dot token from a member access expression</param>
+    /// <param name="dotToken">The dot or null-forgiving operator token from a chain link</param>
     /// <returns><see langword="true"/> if there is an intermediate member access; otherwise, <see langword="false"/></returns>
     public static bool DotHasIntermediateMemberAccess(SyntaxToken dotToken)
     {
@@ -267,6 +269,12 @@ internal static class ChainWalker
         {
             return memberAccess.Expression is MemberAccessExpressionSyntax
                    || memberAccess.Expression is ConditionalAccessExpressionSyntax;
+        }
+
+        if (dotToken.Parent is PostfixUnaryExpressionSyntax postfixUnary)
+        {
+            return postfixUnary.Operand is MemberAccessExpressionSyntax
+                   || postfixUnary.Operand is ConditionalAccessExpressionSyntax;
         }
 
         return false;
--- a/Reihitsu.Analyzer.Test/Formatting/RH5112WrappedFluentCallsShouldKeepFirstCallOnOriginalLineAnalyzerTests.cs
+++ b/Reihitsu.Analyzer.Test/Formatting/RH5112WrappedFluentCallsShouldKeepFirstCallOnOriginalLineAnalyzerTests.cs
@@ -315,7 +315,7 @@ public class RH5112WrappedFluentCallsShouldKeepFirstCallOnOriginalLineAnalyzerTe
     /// <summary>
     /// Verifies that RH5112 does not report when the first invoked link is introduced by a
     /// null-forgiving operator whose own receiver is an intermediate member access, matching the
-    /// existing exemption for a plain dot in the same position (issue #721 review)
+    /// existing exemption for a plain dot in the same position (PR #721)
     /// </summary>
     /// <returns>A <see cref="Task"/> representing the asynchronous operation</returns>
     [TestMethod]
@@ -355,6 +355,68 @@ public class RH5112WrappedFluentCallsShouldKeepFirstCallOnOriginalLineAnalyzerTe
                                 """;
 
         await Verify(testData);
+    }
+
+    /// <summary>
+    /// Verifies the other side of the intermediate-member-access boundary: RH5112 still reports, and
+    /// its code fix still converges, when the first invoked link is introduced by a null-forgiving
+    /// operator whose own receiver is <em>not</em> an intermediate member access (PR #721)
+    /// </summary>
+    /// <returns>A <see cref="Task"/> representing the asynchronous operation</returns>
+    [TestMethod]
+    public async Task VerifyDiagnosticAndCodeFixWhenNullForgivingLinkHasNoIntermediateMemberAccess()
+    {
+        const string testData = """
+                                internal sealed class Example
+                                {
+                                    private static object Create(Builder builder)
+                                    {
+                                        return builder
+                                            {|#0:!|}.UseLogging()
+                                            .Build();
+                                    }
+
+                                    private sealed class Builder
+                                    {
+                                        public Builder UseLogging()
+                                        {
+                                            return this;
+                                        }
+
+                                        public object Build()
+                                        {
+                                            return new object();
+                                        }
+                                    }
+                                }
+                                """;
+        const string resultData = """
+                                  internal sealed class Example
+                                  {
+                                      private static object Create(Builder builder)
+                                      {
+                                          return builder!.UseLogging()
+                                                        .Build();
+                                      }
+
+                                      private sealed class Builder
+                                      {
+                                          public Builder UseLogging()
+                                          {
+                                              return this;
+                                          }
+
+                                          public object Build()
+                                          {
+                                              return new object();
+                                          }
+                                      }
+                                  }
+                                  """;
+
+        await Verify(testData,
+                     resultData,
+                     Diagnostics(RH5112WrappedFluentCallsShouldKeepFirstCallOnOriginalLineAnalyzer.DiagnosticId, AnalyzerResources.RH5112MessageFormat));
     }
 
     #endregion // Tests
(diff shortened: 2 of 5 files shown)


Reihitsu.Formatter/Pipeline/LineBreaks/Utilities/ChainWalker.cs
```diff
@@ -269,6 +271,12 @@ public static bool DotHasIntermediateMemberAccess(SyntaxToken dotToken)
                    || memberAccess.Expression is ConditionalAccessExpressionSyntax;
         }
```
thoenissen, line 272: This new postfix-unary branch has no counterpart in `Reihitsu.Analyzer/Core/FluentChainAnalysisHelper.cs`'s `HasIntermediateMemberAccess`, which is the analyzer-side copy of the same predicate (used by RH5112) and still only switches on `MemberAccessExpressionSyntax`/`ConditionalAccessExpressionSyntax`. This creates a confirmed parity break and a non-convergent code fix for a null-forgiving operator wrapped directly in front of an invoked link whose own receiver has an intermediate member access:

```csharp
var x = builder.Inner
    !.UseLogging()
    .Build();
```

Verified by execution on this PR's head: RH5112 reports a diagnostic here (`chainLinks[0]` resolves to the `!` token via `FluentChainUtilities.GetInvokedLinkOperator`, and the analyzer's `HasIntermediateMemberAccess(!)` falls through its `default` arm to `false`), while the formatter (with this PR's fix) correctly leaves the chain wrapped — `DotHasIntermediateMemberAccess(!)` now returns `true` via this new branch. Because RH5112's code fix (`RH5112WrappedFluentCallsShouldKeepFirstCallOnOriginalLineCodeFixProvider`) delegates to `ReihitsuFormatter.FormatNodeInDocumentAsync`, applying the offered fix does not change the document and the diagnostic persists — a registered code action that cannot change the document. `FluentChainAnalysisHelper.HasIntermediateMemberAccess` needs the same postfix-unary check this PR adds here (or both should move onto one shared `Reihitsu.Core` implementation).

---

thoenissen (commented):
1. `Reihitsu.Formatter.Test/Regression/Indentation/MethodChainAlignmentTests.cs`: the new candidate class this PR introduces — a wrapped null-forgiving operator in the join gap between the chain root and the operator itself (`a` ⏎ `!.Prop...`) — has a comment-gap regression test but no preprocessor-directive-gap test, unlike the sibling plain-dot shape (`PragmaAboveFirstInvokedLinkKeepsEveryInvokedLinkOnItsOwnLine`, `DisabledTextAboveFirstInvokedLinkKeepsEveryInvokedLinkOnItsOwnLine`). `WouldJoinAcrossUnjoinableTrivia` is token-kind-agnostic so this is unlikely to fail, but per the "Test expectations" requirement for any change touching a token-gap join, a `#pragma`/`#if false` variant of `WrappedNullForgivingNonInvokedPrefixCollapsesOntoChainRoot` should be added.

---

Reihitsu.Formatter/Pipeline/LineBreaks/Utilities/ChainWalker.cs
thoenissen: Addressed: `FluentChainAnalysisHelper.HasIntermediateMemberAccess` now has the matching `PostfixUnaryExpressionSyntax` branch, closing the parity break — RH5112 no longer fires (and its fix no longer no-ops) for `builder.Inner` ⏎ `!.UseLogging()` ⏎ `.Build();` (1ee5770). A subsequent official preflight audit found the underlying candidate-search change itself needed narrowing to avoid an idempotency/parity regression on a related shape; that repair (8dd3289) also closes the class this finding is part of, and both sides of the boundary are now pinned by regression tests on both the formatter and RH5112 sides.

---

thoenissen:
Addressed: added `PragmaAboveWrappedNullForgivingPrefixKeepsEveryInvokedLinkOnItsOwnLine` and `DisabledTextAboveWrappedNullForgivingPrefixKeepsEveryInvokedLinkOnItsOwnLine` to `MethodChainAlignmentTests.cs`, mirroring the existing plain-dot sibling tests for the new null-forgiving-prefix candidate class (1ee5770).

---

